PR #1, opened 2020-01-21
Bootstrap operator

Kidswiss:
This commit contains the first versions of the various CRDs and some
logic for the controllers.


srueg:
The built binary should probably go into the `.gitignore`. It is committed now `build/_output/bin/lieutenant-operator-local`

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,75 @@
...
+// FactValue is the value for the facts map
+type FactValue string
+
+// Facts is a map of arbitrary facts for the cluster
+type Facts map[FactKey]FactValue
+
+// ClusterSpec defines the desired state of Cluster
+type ClusterSpec struct {
+	// APIEndpointSecretRef references the secret containing connection information to Kubernetes API endpoint of the registered Kubernetes cluster.
+	APIEndpointSecretRef *corev1.SecretKeySelector `json:"apiEndpointSecretRef,omitempty"`
```
srueg: A `SecretKeySelector` only selects a single key of a secret, instead of the whole secret.
```suggestion
	APIEndpointSecretRef *corev1.SecretReference `json:"apiEndpointSecretRef,omitempty"`
```

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// DisplayName of cluster which could be different from metadata.name. Allows cluster renaming should it be needed.
+	DisplayName string `json:"displayName,omitempty"`
+	// Slug is identification of the cluster (shortname or slug).
+	Slug string `json:"slug,omitempty"`
+	// GitRepoURL git repository storing the cluster configuration catalog. If this is set, no gitRepoTemplate is needed.
+	GitRepoURL string `json:"gitRepoURL,omitempty"`
+	// GitRepoTemplate template for managing the GitRepo object.
+	GitRepoTemplate GitRepoTemplate `json:"gitRepoTemplate,omitempty"`
```
srueg: What happens if it is not specified? Maybe a pointer type makes more sense?

pkg/helpers/crd.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/go-logr/logr"
+	synv1alpha1 "github.com/projectsyn/lieutenant-operator/pkg/apis/syn/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//GitRepoCreator creates git repo objects including correct owner and tenant refs
+type GitRepoCreator struct {
```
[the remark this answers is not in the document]
Kidswiss: Will resolve

pkg/apis/syn/v1alpha1/cluster_types.go
Kidswiss: Assumes from SDD -> should be specified there as well

PR #2, opened 2020-01-24
Finalising cluster logic and add unit tests

Kidswiss:
This commit finalises the first draft for the cluster controller logic
and adds unit tests to all controllers.


deploy/crds/syn.tools_v1alpha1_gitrepo_cr.yaml
```diff
@@ -3,5 +3,7 @@ kind: GitRepo
 metadata:
   name: example-gitrepo
 spec:
-  # Add fields here
-  size: 3
+  apiSecretRef:
+      name: test
```
srueg:
```suggestion
    name: test
```

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -103,3 +103,33 @@ func (r *ReconcileCluster) updateStatus(cluster *synv1alpha1.Cluster) error {
...
+	_, err := r.fetchInstance(cluster, types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace})
+	if err != nil {
+		return err
+	}
+
+	if cluster.ObjectMeta.Labels == nil {
+		cluster.ObjectMeta.Labels = make(map[string]string)
+	}
+	cluster.ObjectMeta.Labels["tenant"] = cluster.Spec.TenantRef.Name
+	return r.client.Update(context.TODO(), cluster)
```
srueg: I'd recommend to only update the object once at the end of the reconcile run to avoid multiple calls to the API and often re-queueing of the object.

pkg/controller/cluster/cluster_reconcile_test.go
```diff
@@ -0,0 +1,179 @@
...
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+
+	_, err := c.Reconcile(req)
+	if err != nil {
+		t.Fatalf("reconcile: (%v)", err)
```
srueg: As I understood it, you'd use `t.Errorf()` instead.
Also the https://github.com/stretchr/testify package could come in handy to use `assert.NoError(t, err)` for example.

srueg:
Very cool how these tests work! :+1:

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: See todos in code.

pkg/controller/cluster/cluster_reconcile_test.go
Kidswiss: Will have a look at testify

Kidswiss: And also try to do the tests table driven.

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: I can't currently set the url in gitRepoURL, as the logic nessary for that will be implemented in SYn-25. So currently it will set it to "dummy".

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -103,3 +116,27 @@ func (r *ReconcileCluster) updateStatus(cluster *synv1alpha1.Cluster) error {
...
+
+func (r *ReconcileCluster) fetchInstance(instance *synv1alpha1.Cluster, namespacedName types.NamespacedName) (reconcile.Result, error) {
+
+	err := r.client.Get(context.TODO(), namespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, err
```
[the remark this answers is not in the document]
srueg: Did you fix this on the other branch?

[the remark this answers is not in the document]
Kidswiss: Yes that is already fixed on the SYN-25 branch.

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -2,31 +2,22 @@ package v1alpha1
...
 	// DisplayName of cluster which could be different from metadata.name. Allows cluster renaming should it be needed.
-	DisplayName string `json:"displayName,omitempty"`
+	DisplayName string `json:"displayName"`
 	// GitRepoURL git repository storing the cluster configuration catalog. If this is set, no gitRepoTemplate is needed.
 	GitRepoURL string `json:"gitRepoURL,omitempty"`
 	// GitRepoTemplate template for managing the GitRepo object.
 	GitRepoTemplate *GitRepoTemplate `json:"gitRepoTemplate,omitempty"`
 	// TenantRef reference to Tenant object the cluster belongs to.
-	TenantRef TenantRef `json:"tenantRef,omitempty"`
+	TenantRef *corev1.LocalObjectReference `json:"tenantRef"`
```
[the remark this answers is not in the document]
Kidswiss: Fine by me

pkg/apis/syn/v1alpha1/gitrepo_types.go
```diff
@@ -37,15 +37,15 @@ func (g GitType) IsValid() bool {
...
+	DeployKeys map[string]DeployKey `json:"deployKeys,omitempty"`
 	// Path to Git repository
-	Path string `json:"path,omitempty"`
+	Path string `json:"path"`
 	// RepoName ame of Git repository
-	RepoName string `json:"repoName,omitempty"`
+	RepoName string `json:"repoName"`
 	// TenantRef references the tenant this repo belongs to
-	TenantRef *TenantRef `json:"tenantRef,omitempty"`
+	TenantRef *corev1.LocalObjectReference `json:"tenantRef,omitempty"`
```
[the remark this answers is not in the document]
Kidswiss: The problem if we declare this one as mandatory is, that you'll also have to add the tenant ref to the gitRepoTemplate which leads to a very weird user experience when defining clusters. Especially for the cluster where you'll end up defining the same tenantRef twice...

One possible solution for that: we could split the spec into a template and non-template part, where the template doesn't contain the tenant ref and the spec does. That way we can  have both cases.

```suggestion
	// GitRepoSpec defines the desired state of GitRepo
type GitRepoSpec struct {
	GitRepoTemplate
	// TenantRef references the tenant this repo belongs to
	TenantRef corev1.LocalObjectReference `json:"tenantRef"`
}

// GitRepoTemplate is used for templating git repos, it does not contain the tenantRef as it will be added by the
// controller creating the template instance.
type GitRepoTemplate struct {
	// APISecretRef reference to secret containing connection information
	APISecretRef corev1.SecretReference `json:"apiSecretRef"`
	// DeployKeys optional list of SSH deploy keys. If not set, not deploy keys will be configured
	DeployKeys map[string]DeployKey `json:"deployKeys,omitempty"`
	// Path to Git repository
	Path string `json:"path"`
	// RepoName ame of Git repository
	RepoName string `json:"repoName"`
}
```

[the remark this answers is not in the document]
Kidswiss: Applied and tested this. It's still a bit weird to handle, but definitely less weird, what do you think?

PR #4, opened 2020-02-07
Manage git repositories via CRDs

Kidswiss:
This commit contains the implementation for Gitlab to manage the repos.


Kidswiss:
@srueg I renamed `bootstrapTokenValid` in the cluster status to `tokenValid` is that okay for you?

Kidswiss:
The git functionality should now be okay for a first version.

Also a proposal:

Let's use golang's ldflags to automagically set the version and build date during CI/CD
```
 go run -ldflags="-X 'main.version=$(git describe --tags)' -X 'main.date=$(date)'" cmd/manager/main.go

# in main.go
// set these with ldflags
var version = "dev"
var date = "now"

func printVersion() {
	log.Info(fmt.Sprintf("Operator Version: %s", version))
	log.Info(fmt.Sprintf("Build Date: %s", date))
*SNIP*
```

srueg:
> The git functionality should now be okay for a first version.
> 
> Also a proposal:
> 
> Let's use golang's ldflags to automagically set the version and build date during CI/CD
> 
> ```
> # set version during build/run
> go build/run -ldflags="-X 'main.version=$(git describe --tags)' -X 'main.date=$(date)'" cmd/manager/main.go
> 
> # in main.go
> // set these with ldflags
> var version = "dev"
> var date = "now"
> 
> func printVersion() {
> 	log.Info(fmt.Sprintf("Operator Version: %s", version))
> 	log.Info(fmt.Sprintf("Build Date: %s", date))
> *SNIP*
> ```

:+1: Implemented in https://github.com/projectsyn/lieutenant-operator/pull/5

pkg/git/gitlab/gitlab.go
```diff
@@ -0,0 +1,295 @@
...
+	for k, v := range localKeys {
+		mergedKey := v.Type + " " + v.Key
+		keyOpts := &gitlab.AddDeployKeyOptions{
+			Title:   &k,
+			Key:     &mergedKey,
+			CanPush: &v.WriteAccess,
+		}
+
+		if force {
+			g.overwriteKey(existingKeys, k)
```
srueg: Maybe we should delete the old key only once adding the new one was successful?

pkg/git/manager/manager.go
```diff
@@ -0,0 +1,76 @@
...
+	Read() error
+	Delete() error
+	Connect() error
+}
+
+// Implementation is a set of functions needed to get the right git implementation
+// for the given URL.
+type Implementation interface {
+	// IsType returns true, if the given URL is handleable by the given implementation (Github,Gitlab, etc.)
+	IsType(URL string) bool
```
srueg: Maybe we should differentiate between false and error, to make debugging easier?
I.e. if a connection error occurs, it'll now look like it couldn't find a proper implementation...

Kidswiss:
@sruegg ready for another look 👍

pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -18,18 +26,100 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	gitRepo.Status.Phase = &phaseToSet
+
+	if phaseToSet == synv1alpha1.Created {
+		gitRepo.Status.URL = repoImpl.FullURL().String()
+	}
+
+	return r.client.Status().Update(context.TODO(), gitRepo)
+}
+
+func (r *ReconcileGitRepo) fetchInstance(instance *synv1alpha1.GitRepo, namespacedName types.NamespacedName) error {
```
srueg: This method can be removed, since only used once

pkg/helpers/crd.go
```diff
@@ -23,11 +23,13 @@ func CreateGitRepo(obj metav1.Object, gvk schema.GroupVersionKind, template *syn
 
 	tenantName := obj.GetName()
 	tenantNamespace := obj.GetNamespace()
-	if tenantRef != nil {
+	if tenantRef != nil || tenantName == "" {
 		tenantName = tenantRef.Name
```
srueg: `tenantRef` could be `nil` here. Should probably be AND instead of OR?

srueg:
For the `GitRepoSpec`, could we just embed the `GitRepoTemplate` instead of having a property with the name? It would reduce the nesting (by removing `gitRepoTemplate:`).

Kidswiss:
> For the `GitRepoSpec`, could we just embed the `GitRepoTemplate` instead of having a property with the name? It would reduce the nesting (by removing `gitRepoTemplate:`).

It's already implemented like this:
```
// GitRepoSpec defines the desired state of GitRepo
type GitRepoSpec struct {
	GitRepoTemplate `json:"gitRepoTemplate"`
	// TenantRef references the tenant this repo belongs to
	TenantRef corev1.LocalObjectReference `json:"tenantRef"`
}
```

Although I had to change the json tag from `json:"gitRepoTemplate"` to `json:",inline"` to make the code generator happy again, is it requires a valis json tag for every field. Now the additional nesting isn't necessary anymore.

pkg/helpers/crd.go
Kidswiss: Simplified the logic.

pkg/apis/syn/v1alpha1/gitrepo_types.go
```diff
@@ -63,31 +55,21 @@ type DeployKey struct {
...
-	Type               string      `json:"type,omitempty"`
-}
-
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // GitRepo is the Schema for the gitrepos API
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=gitrepos,scope=Namespaced
-// +kubebuilder:printcolumn:name="Repo Name",type="string",JSONPath=".spec.repoName"
+// +kubebuilder:printcolumn:name="Repo Name",type="string",JSONPath=".spec.gitRepoTemplate.repoName"
```
srueg:
```suggestion
// +kubebuilder:printcolumn:name="Repo Name",type="string",JSONPath=".spec.repoName"
```

PR #5, opened 2020-02-07
GitHub actions

srueg:


Kidswiss:
LGTM

I'm rebasing this onto SYN-25 so we'll have the actions ready there.

PR #11, opened 2020-02-12
WIP Fix path

Kidswiss:


pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -66,11 +66,12 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 		},
 		DeployKeys: instance.Spec.DeployKeys,
 		Logger:     reqLogger,
+		BaseURL:    string(secret.Data[SecretEndpointName]),
```
srueg: Let's use a `URL` instead of string here. Also check if the secrets contains a `endpoint` and `token`.

pkg/git/gitlab/gitlab.go
```diff
@@ -130,7 +125,7 @@ func (g *Gitlab) getProject() error {
 
 	// we need to remove the leading slash from the path or else the HTML encode
 	// for the getproject call will fail...
-	path := strings.Replace(g.url.Path, "/", "", 1)
+	path := strings.Replace(g.ops.Path, "/", "", 1)
```
srueg: This shouldn't be necessary anymore

pkg/git/manager/manager.go
```diff
@@ -45,6 +45,9 @@ type RepoOptions struct {
 	Credentials Credentials
 	DeployKeys  map[string]synv1alpha1.DeployKey
 	Logger      logr.Logger
+	BaseURL     string
```
srueg: URL instead of string

PR #13, opened 2020-02-13
More integration fixes

Kidswiss:


Kidswiss:
@srueg this is ready for reviewing

pkg/helpers/crd.go
```diff
@@ -51,16 +52,21 @@ func CreateGitRepo(obj metav1.Object, gvk schema.GroupVersionKind, template *syn
...
+
+		return err
 
-		return false, client.Update(context.TODO(), existingRepo)
 	} else if err != nil {
-		return false, err
+		return err
 	}
-	return true, nil
+	return nil
```
srueg: You can collapse all the preceding returns into one `return err` and also remove the `else if`

PR #22, opened 2020-02-17
Implement GitRepo type

srueg:
This value specifies if a GitRepo object is actually handled by the git
controller. A value of `unmanaged` means the controller won't manage it
and it's sole purpose is data storage. An empty value defaults to `auto`
which means it's managed by the controller.

Signed-off-by: Simon Rüegg <simon.ruegg@vshn.ch>

### DIFF
--- a/pkg/controller/gitrepo/gitrepo_reconcile.go
+++ b/pkg/controller/gitrepo/gitrepo_reconcile.go
@@ -42,6 +42,9 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 	helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
+	if instance.Spec.RepoType == synv1alpha1.DefaultRepoType {
+		instance.Spec.RepoType = synv1alpha1.AutoRepoType
+	}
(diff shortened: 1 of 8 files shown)


pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -42,6 +42,9 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 	helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
+	if instance.Spec.RepoType == synv1alpha1.DefaultRepoType {
```
Kidswiss: It's possible to set defaults via markers directly in the CRD struct: https://book.kubebuilder.io/reference/markers/crd-validation.html

So we don't have to clutter the logic with such if statements.

srueg: Unfortunately this only seems to be GA since K8s v1.17: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting

I guess we should leave the code there for now?

Kidswiss: That's unfortunate but to support older clusters let's leave the code.

Issue #37, opened 2020-03-30
Write Documentation


srueg:
Write some basic documentation about this tool. Reuse as much as possible from the SDD: https://github.com/projectsyn/documentation/blob/master/docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
If possible, generate documentation for the OpenAPI CRDs (e.g. using [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs)) and include it in the Antora docs.

## Task Deliverables

- [ ] Basic documentation about lieutenant-operator

Kidswiss:
I added a bit of a hacky script to`generate.go` which uses your linked generator to generate a html file for our CRDs. That file in turn is included in the asciidoc documentation.

I also transferred everything from the SDD to this repo. The CRD topics are now a bit redundant. Should I remove the SDD CRD definitions in favour of the concrete implementations? @srueg

srueg:
Nice! Yes good point. I think only leaving the generated part is ok.

Kidswiss:
Implemented your feedback and removed the old CRD documentation.

Kidswiss:
Unfortunately the html include does not work currently: https://docs.syn.tools/lieutenant-operator/crds-html.html

Will investigate

srueg:
@akosma can you please add the Makefile setup for Antora to this repo (and also include it in the GitHub action)?

Kidswiss:
After quite some investigation I found two things:
* the include has to be done in a special manner for Antora to pick it up
* I needed to adjust the HTML templates used to generate the CRD docs so it will get the right styling

Lokal renders look good now:
![image](https://user-images.githubusercontent.com/7460550/79337167-52ecf580-7f25-11ea-825c-08bea10fd4bc.png)

@srueg Can you have a look and merge if everything looks good to you? THX

srueg:
Looking good 👍 https://syn.tools/lieutenant-operator/crds-html.html

Kidswiss:
Nice!

The docs now look good, closing here.

PR #40, opened 2020-04-09
Docs

Kidswiss:
Refers to: #37


create-api-docs.sh
```diff
@@ -0,0 +1,23 @@
...
+orig_path=$(pwd)
+
+export GOPATH="$(pwd)/tempgopath"
+export GO111MODULE="on"
+
+mkdir -p "$GOPATH/src/github.com/projectsyn"
+
+ln -s "$(pwd)" "$GOPATH/src/github.com/projectsyn/lieutenant-operator"
+
+cd "$GOPATH/src/github.com/projectsyn/lieutenant-operator" || return 1
```
srueg, line 14: Could the `return` be done instead by using `set -euo pipefail` at the beginning?

Kidswiss: Yes, that should have the same effect. Adding that.

Issue #36, opened 2020-03-30
Implement Tenant GitRepo Bootstrapping


srueg:
Automate creation of cluster config file in tenant configuration repository.
For each new cluster, an empty YAML file needs to be created in the tenant's config repo. The file **must** be called `$CLUSTER_ID.yml` (note the `.yml` **without** `a`)

## Task Deliverables

- [ ] Operator automatically creates `$CLUSTER_ID.yml` files in the tenant repo

Kidswiss:
Started to implement this in branch `git/bootstrapping`.

I'm implementing some generalised bootstrap handling so that we can commit an arbitrary amount of files into the repository.

Kidswiss:
Commiting files into a repository is now implemented.

The CRD for the GitRepository was extended in a way that it can hold a map of file/content pairs. Each of these entries will be committed to the repository upon the next reconcile loop.

I also found and fixed two annoying bugs that cost me way to much time to find.

Still to do:

- [ ] unit test where ever possible

Kidswiss:
There are now unit tests for the new functionality. While implementing them I already found a minor bug that I fixed.

@srueg would you have a look at #42 ?

THX

Kidswiss:
Changes merged, closing here.

PR #42, opened 2020-04-15
Implement git file templates

Kidswiss:
With this commit the GitRepo CRD was extended so it can hold a map of
file and content. Currently it's only tested to create files in the
root of a repository and not deeper in the file tree.
    
Also each cluster will now create an empty template file in its tenant
repository.
    
Relates to: #36

### DIFF
--- a/pkg/apis/syn/v1alpha1/gitrepo_types.go
+++ b/pkg/apis/syn/v1alpha1/gitrepo_types.go
@@ -54,7 +54,7 @@ type GitRepoTemplate struct {
 	DeployKeys map[string]DeployKey `json:"deployKeys,omitempty"`
 	// Path to Git repository
 	Path string `json:"path,omitempty"`
-	// RepoName ame of Git repository
+	// RepoName name of Git repository
 	RepoName string `json:"repoName,omitempty"`
 	// RepoType specifies if a repo should be managed by the git controller. A value of 'unmanaged' means it's not manged by the controller
 	// +kubebuilder:validation:Enum=auto;unmanaged
--- a/pkg/apis/syn/v1alpha1/gitrepo_types.go
+++ b/pkg/apis/syn/v1alpha1/gitrepo_types.go
@@ -49,16 +49,19 @@ type GitRepoSpec struct {
 // controller creating the template instance.
 type GitRepoTemplate struct {
 	// APISecretRef reference to secret containing connection information
-	APISecretRef corev1.SecretReference `json:"apiSecretRef"`
+	APISecretRef corev1.SecretReference `json:"apiSecretRef,omitempty"`
 	// DeployKeys optional list of SSH deploy keys. If not set, not deploy keys will be configured
 	DeployKeys map[string]DeployKey `json:"deployKeys,omitempty"`
 	// Path to Git repository
-	Path string `json:"path"`
+	Path string `json:"path,omitempty"`
 	// RepoName ame of Git repository
-	RepoName string `json:"repoName"`
+	RepoName string `json:"repoName,omitempty"`
 	// RepoType specifies if a repo should be managed by the git controller. A value of 'unmanaged' means it's not manged by the controller
 	// +kubebuilder:validation:Enum=auto;unmanaged
 	RepoType RepoType `json:"repoType,omitempty"`
+	// TemplateFiles is a list of files that should be pushed to the repository
+	// after its creation.
+	TemplateFiles map[string]string `json:"templateFiles,omitempty"`
 }
 
 // DeployKey defines an SSH key to be used for git operations.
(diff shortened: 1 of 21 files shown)


pkg/apis/syn/v1alpha1/gitrepo_types.go
```diff
@@ -49,16 +49,19 @@ type GitRepoSpec struct {
...
 type GitRepoTemplate struct {
 	// APISecretRef reference to secret containing connection information
-	APISecretRef corev1.SecretReference `json:"apiSecretRef"`
+	APISecretRef corev1.SecretReference `json:"apiSecretRef,omitempty"`
 	// DeployKeys optional list of SSH deploy keys. If not set, not deploy keys will be configured
 	DeployKeys map[string]DeployKey `json:"deployKeys,omitempty"`
 	// Path to Git repository
-	Path string `json:"path"`
+	Path string `json:"path,omitempty"`
 	// RepoName ame of Git repository
```
srueg, line 57:
```suggestion
	// RepoName name of Git repository
```

pkg/apis/syn/v1alpha1/gitrepo_types.go
```diff
@@ -49,16 +49,19 @@ type GitRepoSpec struct {
 // controller creating the template instance.
 type GitRepoTemplate struct {
 	// APISecretRef reference to secret containing connection information
-	APISecretRef corev1.SecretReference `json:"apiSecretRef"`
+	APISecretRef corev1.SecretReference `json:"apiSecretRef,omitempty"`
```
srueg, line 52: Why `omitempty` (optional)?

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -108,4 +117,27 @@ func (r *ReconcileCluster) newStatus(cluster *synv1alpha1.Cluster) error {
...
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		} else {
+			return nil
+		}
+	}
+
+	if _, ok := tenantCR.Spec.GitRepoTemplate.TemplateFiles[filename]; !ok {
+		if tenantCR.Spec.GitRepoTemplate.TemplateFiles == nil {
```
srueg, line 133: Can this even be `nil` if you just accessed it?

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -108,4 +117,27 @@ func (r *ReconcileCluster) newStatus(cluster *synv1alpha1.Cluster) error {
...
+			tenantCR.Spec.GitRepoTemplate.TemplateFiles = map[string]string{
+				filename: "",
+			}
+		} else {
+			tenantCR.Spec.GitRepoTemplate.TemplateFiles[filename] = ""
+		}
+	}
+
+	return r.client.Update(context.TODO(), tenantCR)
```
srueg, line 142: You'd only need to call `update` if you changed the `tenantCR`...

pkg/helpers/crd.go
```diff
@@ -41,6 +42,10 @@ func CreateOrUpdateGitRepo(obj metav1.Object, gvk schema.GroupVersionKind, templ
 		},
 	}
 
+	// the git repository may not be available right after the creation. So let's give k8s a bit of time to
+	// make it available.
+	defer time.Sleep(10 * time.Millisecond)
```
srueg, line 47: This look suspicious. Is it really necessary?

pkg/apis/syn/v1alpha1/gitrepo_types.go
Kidswiss: If the GitRepo is not managed by the operator, e.g. RepoType=unamanged, you should not be required to add an API key.

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: Accessing a nil map doesn't panic, writing to one does:
https://play.golang.org/p/E77-4k1MPs-

That's a rather weird Golang quirk I found and caused nil pointer panics during testing.

pkg/helpers/crd.go
Kidswiss: I was able to produce "NotFound" errors quite consistently without this wait. Not sure what causes that issue though...

It only happens with Tenants, too, as between `CreateOrUpdateGitRepo()` and `GetGitRepoURLAndHostKeys()` are no other statements.

But I could also implement the `RetryOnConflict()` I used for the GitRepo reconcile.

Kidswiss: Using `retry.OnError()` seems to work fine for this use case.

pkg/controller/cluster/cluster_reconcile.go
srueg: Hmm, weird indeed.
Then I'd suggest to untangle this nested if/else a little: Check if map is `nil` and instantiate an empty one if so. The other if can then be simpler.

Kidswiss: Untangled the nested IF statement.

Issue #33, opened 2020-03-30
Write Cluster Token to Vault


srueg:
The operator should store the cluster service account token in Vault automatically.

## Task Deliverables

- [ ] Service account token of clusters are stored in vault
- [ ] The tokens are stored under `kv/$TENANT_ID/$CLUSTER_ID/steward token=$CLUSTER_TOKEN`

Kidswiss:
As discussed in the SYN meeting: I'll explore the possibilities of k8s token auth.

Kidswiss:
Implemented Vault connection, tested with local Vault and unit tests.

left to do:

- [ ] test on synfra cluster with k8s authentication

srueg:
We configured Vault with the following policy:
```hcl
path "kv/data/*" {
  capabilities = ["read", "create", "update"]
}
```

```console
vault policy write lieutenant-operator policy.hcl
```

And the following role for the Kubernetes Auth provider:

```console
vault write auth/kubernetes/role/lieutenant-operator bound_service_account_namespaces=lieutenant-prod,lieutenant-dev bound_service_account_names="lieutenant-operator" policies=lieutenant-operator ttl=1h token_max_ttl=12h
```

srueg:
Maybe worth giving a try: https://pkg.go.dev/github.com/banzaicloud/bank-vaults/pkg/sdk@v0.3.0/vault?tab=doc

srueg:
To follow up our discussion regarding ServiceAccount token rotation:

You can add a new token by creating a (empty) secret: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#to-create-additional-api-tokens
The controller will then add a token to the secret.

Unfortunately it seems that the `secrets` array of the ServiceAccount is not updated with this new secret.

Kidswiss:
I can get all secrets, sort by newest creation date, iteraterate and get the newest one that matches the `kubernetes.io/service-account.name` to the cluster service account.

srueg:
Yes, that's pretty much how `kubectl describe serviceaccount <name>` works: https://github.com/kubernetes/kubectl/blob/master/pkg/describe/describe.go#L2771

Kidswiss:
Implemented all the todos we discussed yesterday (except the gitrepo thing).

PR #47, opened 2020-04-23
Feature/vault integration

Kidswiss:
With this commit the cluster service account token will be added to
    vault.

Closes: #33


pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -63,6 +68,21 @@ func (r *ReconcileCluster) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	token, err := r.getServiceAccountToken(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = client.SetToken(path.Join(instance.Spec.TenantRef.Name, instance.Name), token, reqLogger)
```
srueg, line 81: The token should be saved in `<tenantId>/<clusterId>/steward` with `token=<token>`
```suggestion
	err = client.SetToken(path.Join(instance.Spec.TenantRef.Name, instance.Name, "steward"), token, reqLogger)
```

pkg/vault/client.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"path"
+
+	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
+	"github.com/go-logr/logr"
+	"github.com/hashicorp/vault/api"
+)
+
+const (
+	tokenName    = "steward token"
```
srueg, line 13:
```suggestion
	tokenName    = "token"
```

pkg/vault/client.go
```diff
@@ -0,0 +1,100 @@
...
+	k8sTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	tokenEnvVar  = "VAULT_TOKEN"
```
srueg, line 15: We could reuse `api.EnvVaultToken` from the Vault client

pkg/vault/client.go
```diff
@@ -0,0 +1,100 @@
...
+	vaultEnvVar  = "VAULT_URL"
```
srueg, line 16: We could reuse `api.EnvVaultAddress` from the Vault client

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -141,3 +161,24 @@ func (r *ReconcileCluster) updateTenantGitRepo(tenant types.NamespacedName, file
...
+	if err != nil {
+		return "", err
+	}
+
+	sortSecrets := helpers.SecretSortList(*secrets)
+
+	sort.Sort(sort.Reverse(sortSecrets))
+
+	for _, secret := range sortSecrets.Items {
+		if secret.Annotations[corev1.ServiceAccountNameKey] == instance.Name {
```
srueg, line 178: Let's also check for `secret.Type == corev1.SecretTypeServiceAccountToken` and if the secret really has a key `secret.Data["token"]` (if a new secret is created, it doesn't have the token yet)...

pkg/controller/cluster/cluster_reconcile_test.go
```diff
@@ -85,6 +87,17 @@ func TestReconcileCluster_Reconcile(t *testing.T) {
 						GitRepoTemplate: &synv1alpha1.GitRepoTemplate{},
 					},
 				},
+				&corev1.Secret{
```
srueg, line 90: Would be nice to have a test case with multiple secrets, to test the sorting

pkg/vault/client.go
```diff
@@ -0,0 +1,100 @@
...
+	secret, err := b.client.RawClient().Logical().Read(queryPath)
+	if err != nil {
+		return err
+	}
+
+	if secret == nil || secret.Data[tokenName] != token {
+
+		log.WithName("vault").Info("secrets don't match, re-applying")
+
+		data := map[string]interface{}{
```
srueg, line 89: We should probably reuse the `secret` map here to prevent removing other entries than `token`

Kidswiss: While looking into this I found quite the bug in the update logic, where I didn't unmarshal the secret correctly so it was updated on every reconcile-loop. That's now fixed.

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: This returns now a specific error that gets checked in the reconcile function and then ignored. So we don't trigger unnecessary reconciles while waiting for the token to be populated.

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -63,6 +68,23 @@ func (r *ReconcileCluster) Reconcile(request reconcile.Request) (reconcile.Resul
...
 	repoName.Name = instance.Spec.TenantRef.Name
 
+	client, err := vault.NewClient()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	token, err := r.getServiceAccountToken(instance)
+	if err != nil {
+		if err.Error() != "no token available yet" {
```
srueg, line 78: Please create a proper error instead of string comparison

pkg/vault/client.go
```diff
@@ -0,0 +1,105 @@
...
+		log.WithName("vault").Info("does not yet exist, creating", "name", secretPath)
+		secret = &api.Secret{}
+		secret.Data = vault.NewData(0, map[string]interface{}{
+			tokenName: token,
+		})
+		_, err = b.client.RawClient().Logical().Write(queryPath, secret.Data)
+		return err
+	}
+
+	secretData := secret.Data["data"].(map[string]interface{})
```
srueg, line 93: Maybe add a type check?

pkg/vault/client_test.go
```diff
@@ -0,0 +1,77 @@
...
+func testGetHTTPServer(statusCode int, body []byte) *httptest.Server {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		res.WriteHeader(statusCode)
+		_, _ = res.Write(body)
+	}))
+
+	return testServer
+}
+
+func TestNewClient(t *testing.T) {
```
srueg, line 22: Maybe add another test for the `SetToken()` create/update path?

pkg/controller/cluster/cluster_reconcile.go
srueg: Also this should still `return`, right?
Thinking about it, I don't think this should be a special case. It's just an error and should be handled like one...

pkg/controller/cluster/cluster_reconcile_test.go
```diff
@@ -85,6 +85,32 @@ func TestReconcileCluster_Reconcile(t *testing.T) {
...
+						Annotations: map[string]string{
+							corev1.ServiceAccountNameKey: tt.fields.objName,
+						},
+						CreationTimestamp: metav1.Time{Time: time.Now()},
+					},
+					Type: corev1.SecretTypeServiceAccountToken,
+					Data: map[string][]byte{
+						"token": []byte("mysecret1"),
+					},
+				},
```
srueg, line 113: Let's add a test case with a secret and missing `token`.

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: As discussed we simply skip this case and return the first found secret with a token.

pkg/vault/client.go
Kidswiss: As discussed: makes no sense, as the type of the `data` field is defined by the secret engine. If that's not correct we're talking to the wrong engine anyways...

pkg/controller/cluster/cluster_reconcile_test.go
Kidswiss: Added a case where the newest secret doesn't have a token and falls back to the previous one.

pkg/vault/client_test.go
Kidswiss: Added separate test case

PR #50, opened 2020-04-29
Update GitRepo objects properly

srueg:
Update both the object with it's labels and the status. Set the status
to "created" if no errors occured.

Signed-off-by: Simon Rüegg <simon@rueggs.ch>


pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -137,7 +137,12 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 			reqLogger.Info("keys differed from CRD, keys re-applied to repository")
 		}
 
-		helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
```
srueg, line 140: The tenant label is added on line 46: https://github.com/projectsyn/lieutenant-operator/pull/50/files#diff-0665f603134a69d730eba6befa1225f9R46

PR #51, opened 2020-04-29
Fix reconcile loop with GitRepo updates

Kidswiss:
Updating the git repository every on every reconcile triggers a loop.
Not quite sure why it happens though, the diffs of the GitRepo objects
don't show any differences apart from `generation` and `resourceVersion`.

The operator will permanently reconcile all tenants and clusters as they
have the Gitrepository as their secondary objects. That triggers and
endless loop of reconciles.

Resolves: #54

### DIFF
--- a/pkg/controller/gitrepo/gitrepo_reconcile.go
+++ b/pkg/controller/gitrepo/gitrepo_reconcile.go
@@ -28,6 +28,9 @@ const (
 // Reconcile will create or delete a git repository based on the event.
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+
+// ATTENTION: do not manipulate the spec here, this will lead to loops, as the specs are
+// defined in the gitrepotemplates of the other CRDs (tenant, cluster).
 func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling GitRepo")
@@ -43,10 +46,7 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 			return err
 		}
-		helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
-		if instance.Spec.RepoType == synv1alpha1.DefaultRepoType {
-			instance.Spec.RepoType = synv1alpha1.AutoRepoType
-		}
+
 		secret := &corev1.Secret{}
 		namespacedName := types.NamespacedName{
 			Name:      instance.Spec.APISecretRef.Name,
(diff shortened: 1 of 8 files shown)


pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -43,7 +43,12 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 			return err
 		}
-		helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
+		if helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name) {
+			err := r.client.Update(context.TODO(), instance)
```
srueg, line 47: I fear we have to call `Update` further down since we add/change other things later which won't ever be written like that...

Kidswiss: Right, moved it to the bottom of the retry block.

pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -137,9 +137,11 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 			reqLogger.Info("keys differed from CRD, keys re-applied to repository")
 		}
 
-		err = r.client.Update(context.TODO(), instance)
-		if err != nil {
-			return err
+		if helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name) {
+			err := r.client.Update(context.TODO(), instance)
```
srueg, line 141: Now we'll only update a gitRepo if the labels changed. There could be other changes as well that need to be written back...

Kidswiss:
Found the issue:

There's  a race condition between the gitRepo templates of the tenants and clusters and the actual gitRepo objects.

The gitRepo controller adds other fields to the objects, that aren't part of the templates. The previous logic simply replaced the `Spec` from the applied gitRepo object with the `Spec` of the templates. That always generated changes due to the missing fields thus causing the loop.

I added logic so the templates will be merged with the applied objects.

pkg/helpers/crd.go
```diff
@@ -54,22 +57,22 @@ func CreateOrUpdateGitRepo(obj metav1.Object, gvk schema.GroupVersionKind, templ
 		if err != nil {
 			return fmt.Errorf("could not update existing repo: %v", err)
 		}
+		existingRepo.Spec = repo.Spec
 
-		if !reflect.DeepEqual(existingRepo.Spec, repo.Spec) {
-			existingRepo.Spec = repo.Spec
+		err = client.Update(context.TODO(), existingRepo)
```
srueg, line 62: Is the update really necessary? Can't we just leave this to whoever is calling this func?

Kidswiss: Not really sure what the benefit would be by removing the actual update from `CreateOrUpdateGitRepo`. I feel like that would clutter the tenant and cluster reconciles with unnecessary GitRepo code. Also if we implement another CR that needs a GitRepo we'd have to remember to call update after calling `CreateOrUpdateGitRepo`. And removing it here would mean we'd have to add it to two places so we'd have code duplication.

srueg: Sorry, my bad I mixed up the CRs.

Kidswiss: No problem :)

PR #56, opened 2020-05-08
Fix reconcile status

srueg:
Set the URL and GitType for a GitRepo in the reconcile loop.


pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -145,6 +145,8 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 		phase := synv1alpha1.Created
```
Kidswiss, line 146: These few lines are exactly the same as for the creation case (L120-122). It would make sense to move that in a function.

srueg: Was there a reason we `return`ed in the case the repo was newly created?

I removed the return which makes the logic easier.

Kidswiss: Yes makes sense. I can't remember a specific reason. It's probably something from before `handleRepoError` was introduced.

PR #57, opened 2020-05-14
Tenant Common

srueg:
Automatically set up a `common.yml` class in each tenant which is included by each cluster of a tenant.

Add option to disable Vault for local development.

### DIFF
--- a/pkg/controller/tenant/tenant_reconcile.go
+++ b/pkg/controller/tenant/tenant_reconcile.go
@@ -13,6 +13,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// CommonClassName is the name of the tenant's common class
+	CommonClassName = "common"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/controller/tenant/tenant_reconcile.go
```diff
@@ -42,6 +47,14 @@ func (r *ReconcileTenant) Reconcile(request reconcile.Request) (reconcile.Result
 		instance.Spec.GitRepoTemplate.DisplayName = instance.Spec.DisplayName
 	}
 
+	commonClassFile := CommonClassName + ".yml"
+	if instance.Spec.GitRepoTemplate.TemplateFiles == nil {
```
Kidswiss, line 51: This could lead to a race condition with the logic in the `ReconcileCluster()` function, as they both manipulate this map and want to write it back to the K8s API. Could lead to those dreadful `the object has been modified` errors.

We may have to wrap the whole reconcile into a `retry` function to avoid that.

srueg:
I added the env var `SKIP_VAULT_SETUP` to skip the Vault setup in local development.

PR #62, opened 2020-05-27
Make operator sync interval configurable

Kidswiss:
This commit makes thes sync interval for the operator configurable. This
should reduce the load on any external components (git, vault, etc.).

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -34,6 +34,10 @@ import (
 	_ "github.com/projectsyn/lieutenant-operator/pkg/git"
 )
 
+const (
+	defaultSyncInterval = 5 * time.Minute
+)
+
 // Change below variables to serve metrics on different host or port.
 var (
 	Version                   = "unreleased"
@@ -96,7 +100,17 @@ func main() {
 		os.Exit(1)
 	}
 
-	duration := 30 * time.Second
+	duration := defaultSyncInterval
+	if str, ok := os.LookupEnv("LIEUTENANT_SYNC_DURATION"); ok {
+		if d, err := time.ParseDuration(str); err != nil {
+			log.Error(err, "failed to parse LIEUTENANT_SYNC_DURATION")
+		} else if d > 30*time.Second {
+			duration = d
+		} else {
+			log.Info("LIEUTENANT_SYNC_DURATION is less than 30s, setting to 30s")
+			duration = 30 * time.Second
+		}
+	}
 
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -96,7 +100,17 @@ func main() {
 		os.Exit(1)
 	}
 
-	duration := 30 * time.Second
+	duration := func(defaultInterval time.Duration) time.Duration {
```
srueg, line 103: Why is this wrapped in a fuction?

Kidswiss: My first idea was to implement this in the `vars()` section, but realised that I can't use the logging there.

You're right here it's not necessary to use a function.

cmd/manager/main.go
```diff
@@ -96,7 +100,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	duration := 30 * time.Second
+	duration := defaultSyncInterval
+	if str, ok := os.LookupEnv("LIEUTENANT_SYNC_DURATION"); ok {
+		if d, err := time.ParseDuration(str); err != nil {
+			log.Error(err, "failed to parse LIEUTENANT_SYNC_DURATION")
+		} else if d > 30*time.Second {
```
srueg, line 107: Lets at least log a warning if the value is less than 30s

cmd/manager/main.go
```diff
@@ -96,7 +100,15 @@ func main() {
...
+			log.Info("LIEUTENANT_SYNC_DURATION is less than 30s, setting to 30s")
```
srueg, line 108: This should go in another `else` and should say something like:
```suggestion
			log.Infof("LIEUTENANT_SYNC_DURATION is less than %s, setting to %s", defaultSyncInterval, defaultSyncInterval)
```

Kidswiss: oh right 🙈  mixed up the compare signs...

cmd/manager/main.go
```diff
@@ -96,7 +100,16 @@ func main() {
...
 
-	duration := 30 * time.Second
+	duration := defaultSyncInterval
+	if str, ok := os.LookupEnv("LIEUTENANT_SYNC_DURATION"); ok {
+		if d, err := time.ParseDuration(str); err != nil {
+			log.Error(err, "failed to parse LIEUTENANT_SYNC_DURATION")
+		} else if d > 30*time.Second {
+			duration = d
+		} else {
+			log.Info("LIEUTENANT_SYNC_DURATION is less than 30s, setting to 30s")
```
srueg, line 110: The default is not 30s
```suggestion
			log.Infof("LIEUTENANT_SYNC_DURATION is less than 30s, setting to %s", defaultSyncInterval)
```

Kidswiss: This is not meant to enforce the default as a minimal value, but rather to enforce more than 30 seconds. We saw that even 30 seconds is not enough, so setting it lower than that doesn't make sense.

Also the `logr` interface does not support `Infof()`.

srueg: Fine by me, but then you need to do `duration = 30 * time.Second`

PR #76, opened 2020-06-17
Deletion

### LINKED ISSUE #32: Implement GitRepo Deletion
Implement the Git repository cleanup process with a retention policy feature (retain, delete) and by leveraging the finalizer pattern for actually deleting repositories.

## Task Deliverables

- [ ] Controller is able to delete managed Git repositories
- [ ] Deletion policy is implemented
- [ ] Finalizer pattern is used for actual deletion process

Kidswiss:
Resolves #32


pkg/helpers/crd.go
```diff
@@ -102,3 +124,89 @@ func (s SecretSortList) Less(i, j int) bool {
...
+		return false, nil
+	}
+
+	annotationValue, exists := instance.GetAnnotations()[DeleteProtectionAnnotation]
+
+	// this is really weird: https://play.golang.org/p/NJkbjN7fCMa
+	var protected bool
+	var err error
+	if exists {
+		protected, err = strconv.ParseBool(annotationValue)
```
srueg, line 152: Will an empty string be true of false?
If I set the annotation `syn.tools/protected-delete=""` I would want it to be true.

pkg/helpers/crd.go
```diff
@@ -102,3 +124,89 @@ func (s SecretSortList) Less(i, j int) bool {
...
+			// We need to reset the error again, so we don't trigger any unwanted side effects...
+			err = nil
+		}
+	} else {
+		protected = false
+	}
+
+	deleted := false
+	if SliceContainsString(instance.GetFinalizers(), finalizerName) {
+		if !protected {
```
srueg, line 165: If it *is* protected, we shouldn't remove the finalizer, since this would delete the object. So this if should probably include the whole function.

pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -126,6 +75,11 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 
 		helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
+		helpers.AddDeletionProtection(instance)
+
+		if !helpers.SliceContainsString(instance.GetFinalizers(), finalizerName) {
```
srueg, line 80: You can skip the if:
> AddFinalizer accepts an Object and adds the provided finalizer if not present

pkg/helpers/values.go
```diff
@@ -0,0 +1,27 @@
+package helpers
+
+import "os"
+
+const (
+	// The top most value here will be the default as ints default to 0
```
srueg, line 6: I don't think that's the case, there's no int here?

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -205,3 +243,32 @@ func (r *ReconcileCluster) getServiceAccountToken(instance metav1.Object) (strin
...
+
+	if _, ok := tenantCR.Spec.GitRepoTemplate.TemplateFiles[fileName]; ok {
+		delete(tenantCR.Spec.GitRepoTemplate.TemplateFiles, fileName)
+		err := r.client.Update(context.TODO(), tenantCR)
+		if err != nil {
+			return err
+		}
+	}
+
+	return repo.RemoveFile(fileName)
```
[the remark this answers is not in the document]
Kidswiss: I was able to produce race conditions the other way round. As the operator re-applied the already deleted file again, because it was in the map.

pkg/helpers/values.go
Kidswiss: Right, that was when I implemented it with a go Enum. Can be removed.

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -24,6 +24,12 @@ type ClusterSpec struct {
 	TokenLifeTime string `json:"tokenLifeTime,omitempty"`
 	// Facts are key/value pairs for statically configured facts
 	Facts *Facts `json:"facts,omitempty"`
+	// DeletionPolicy defines how the external resources should be treated upon CR deletion.
+	// retain: will not delete any external resources
+	// delete: will delete the external resources
+	// archive: will archive the external resources, if it supports that
+	// +kubebuilder:validation:Enum=delete;retain;archive
```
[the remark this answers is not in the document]
Kidswiss: Right there's already a type alias for that. I can re-use it.

pkg/helpers/crd.go
Kidswiss: `syn.tools/protected-delete=""` will throw an error, which is caught and assumed `true` via the if statement on line 154.

pkg/controller/cluster/cluster_reconcile.go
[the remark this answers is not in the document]
srueg: Hmm makes sense. But like this if the `repo.RemoveFile` call fails, it won't be retried.
Anyways it's not really nice to call repo specific stuff from the cluster reconciler. Maybe we should introduce a "deleteMap"? Or set something like `tenantCR.Spec.GitRepoTemplate.TemplateFiles[fileName] = "remove"`?
And then let the repo reconciler handle the actual deletion.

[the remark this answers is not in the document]
Kidswiss: I had similar thoughts while implementing this, but found shortcomings for all my ideas. This one was the "best" one.

Got time for a zoom call today?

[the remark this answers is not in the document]
Kidswiss: As discussed in the zoom call:
* we agreed, that a more complex type to handle this would be the best solution, like:
```
'filename':
  content: 'somestring'
'filename2':
  delete: true
```
* we're at the point where we'd need validation and conversion webhook endpoints
* For the scope of this ticket we'll stick with the deletion magic string so we won't break the API
* I'll open follow-up issues to look into the webhooks

pkg/helpers/crd.go
srueg: Yes, but that's fine in my opinion. It's just how finalizers work. Once we have validating webhooks, we can improve the UX by not deleting the object at all as long as the annotation exists.

Kidswiss: If it won't cause the API to fall over then it's also fine by me. I'll change it tomorrow.

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -24,6 +24,12 @@ type ClusterSpec struct {
 	TokenLifeTime string `json:"tokenLifeTime,omitempty"`
 	// Facts are key/value pairs for statically configured facts
 	Facts *Facts `json:"facts,omitempty"`
+	// DeletionPolicy defines how the external resources should be treated upon CR deletion.
+	// retain: will not delete any external resources
```
srueg, line 28:
```suggestion
	// Retain: will not delete any external resources
```

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -24,6 +24,12 @@ type ClusterSpec struct {
...
+	// delete: will delete the external resources
```
srueg, line 29:
```suggestion
	// Delete: will delete the external resources
```

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -24,6 +24,12 @@ type ClusterSpec struct {
...
+	// archive: will archive the external resources, if it supports that
```
srueg, line 30:
```suggestion
	// Archive: will archive the external resources, if it supports that
```

pkg/apis/syn/v1alpha1/gitrepo_types.go
```diff
@@ -64,6 +70,12 @@ type GitRepoTemplate struct {
 	// TemplateFiles is a list of files that should be pushed to the repository
 	// after its creation.
 	TemplateFiles map[string]string `json:"templateFiles,omitempty"`
+	// DeletionPolicy defines how the external resources should be treated upon CR deletion.
+	// retain: will not delete any external resources
```
srueg, line 74: Please make them uppercase (see suggestions before)

pkg/apis/syn/v1alpha1/tenant_types.go
```diff
@@ -12,6 +12,12 @@ type TenantSpec struct {
 	GitRepoURL string `json:"gitRepoURL,omitempty"`
 	// GitRepoTemplate Template for managing the GitRepo object. If not set, no  GitRepo object will be created.
 	GitRepoTemplate *GitRepoTemplate `json:"gitRepoTemplate,omitempty"`
+	// DeletionPolicy defines how the external resources should be treated upon CR deletion.
+	// retain: will not delete any external resources
```
srueg, line 16: Please make them uppercase (see suggestions before)

pkg/controller/cluster/cluster_reconcile.go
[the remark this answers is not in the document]
Kidswiss: Changed it to use the magic deletion string. After the file has been deleted it will also be removed from the map in the CRD.

I'm now using a type internally that's like the one we discussed in the zoom call. When we change the CRDs we can simplify the logic more.

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -93,10 +107,27 @@ func (r *ReconcileCluster) Reconcile(request reconcile.Request) (reconcile.Resul
...
 
-		err = client.SetToken(path.Join(instance.Spec.TenantRef.Name, instance.Name, "steward"), token, reqLogger)
+		err = vault.SetToken(secretPath, token)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+
+		vault.SetDeletePath(secretPath)
+
+		deleted, err := helpers.HandleDeletion(instance, vault, finalizerName, r.client)
```
srueg, line 117: This error is never handled, if `deleted` is `true`

pkg/helpers/crd.go
```diff
@@ -102,3 +131,85 @@ func (s SecretSortList) Less(i, j int) bool {
...
+// The resource to be removed can be nil, in case there's no direct resource,
+// like for a cluster without Vault.
+func HandleDeletion(instance metav1.Object, resource Remover, finalizerName string, client client.Client) (bool, error) {
+	if instance.GetDeletionTimestamp() == nil {
+		return false, nil
+	}
+
+	annotationValue, exists := instance.GetAnnotations()[DeleteProtectionAnnotation]
+
+	// this is really weird: https://play.golang.org/p/NJkbjN7fCMa
```
srueg, line 157: It's because you shadow `protected` in the scope of the `if`. See https://play.golang.org/p/_e2pkyyzamD

pkg/vault/client.go
```diff
@@ -85,16 +97,99 @@ func (b *BankVaultClient) SetToken(secretPath, token string, log logr.Logger) er
...
+
 		_, err = b.client.RawClient().Logical().Write(queryPath, secret.Data)
 	}
 
 	return err
 }
+
+// SetDeletePath is needed for Remove to work properly. Set the secretPath before
+// calling Remove().
+func (b *BankVaultClient) SetDeletePath(secretPath string) {
```
srueg, line 122: This effectively limits this client to only write/delete one secret in vault. If in the future we want to manage more thsi becomes a problem.

pkg/helpers/crd.go
```diff
@@ -102,3 +131,85 @@ func (s SecretSortList) Less(i, j int) bool {
...
+	return false
+}
+
+// handleDeletion will handle the finalizers if the object was deleted.
+// It will return true, if the deletiontimestamp is set. If the object was
+// removed the reconcile can be returned.
+// The resource to be removed can be nil, in case there's no direct resource,
+// like for a cluster without Vault.
+func HandleDeletion(instance metav1.Object, resource Remover, finalizerName string, client client.Client) (bool, error) {
+	if instance.GetDeletionTimestamp() == nil {
```
srueg, line 151:
```suggestion
	if instance.GetDeletionTimestamp().IsZero() {
```

> IsZero returns true if the value is nil or time is zero

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -93,10 +107,30 @@ func (r *ReconcileCluster) Reconcile(request reconcile.Request) (reconcile.Resul
 			return reconcile.Result{}, err
 		}
 
-		err = client.SetToken(path.Join(instance.Spec.TenantRef.Name, instance.Name, "steward"), token, reqLogger)
+		err = vault.SetToken(secretPath, token)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		vault.SetDeletePaths([]string{secretPath})
```
srueg, line 115: Could we add the functionality of `SetDeletePaths` to `SetToken` so we don't have to call it explicitly?

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -93,10 +107,30 @@ func (r *ReconcileCluster) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		if deleted {
+			err = r.removeClusterFileFromTenant(instance.GetName(), repoName, reqLogger)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
+	}
+
+	deleted, err := helpers.HandleDeletion(instance, nil, finalizerName, r.client)
```
srueg, line 130: Seems kinda weird to call `HandleDeletion` again here. Maybe move it out of the `if` and only call it once?

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -153,11 +190,20 @@ func (r *ReconcileCluster) newStatus(cluster *synv1alpha1.Cluster) error {
 	return nil
 }
 
+func (r *ReconcileCluster) getTenantCR(tenant types.NamespacedName) (*synv1alpha1.Tenant, error) {
+	tenantCR := &synv1alpha1.Tenant{}
+
+	err := r.client.Get(context.TODO(), tenant, tenantCR)
+	if err != nil {
```
srueg, line 197: this `if` is unnecessary since we return `err` anyways, no?

pkg/helpers/crd.go
```diff
@@ -3,18 +3,34 @@ package helpers
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	protectionAnnotation    = "syn.tools/protected-delete"
```
srueg, line 24: Please reuse the const `DeleteProtectionAnnotation `

pkg/helpers/crd.go
```diff
@@ -26,14 +42,27 @@ func CreateOrUpdateGitRepo(obj metav1.Object, gvk schema.GroupVersionKind, templ
...
+			delete(template.TemplateFiles, file)
+		}
+	}
+
 	repo := &synv1alpha1.GitRepo{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      obj.GetName(),
 			Namespace: obj.GetNamespace(),
+			Annotations: map[string]string{
+				DeleteProtectionAnnotation: obj.GetAnnotations()[DeleteProtectionAnnotation],
```
srueg, line 64: If the annotation is not set on `obj` this will return an empty string which in turn is interpreted as `true`

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: I opted for this solution because `SetToken` expects data and will also create the token.

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: Hmmm... if I create a nil vault client before the if this could work.

pkg/helpers/crd.go
Kidswiss: Actually there's already `AddDeletionProtection()` that handles this. Changing it to use that.

pkg/controller/cluster/cluster_reconcile.go
Kidswiss: Tested that on playground: https://play.golang.org/p/RMqmbB4n9vD

I thought tenantCR will still remain `nil` in that case.

pkg/controller/cluster/cluster_reconcile.go
srueg: But the `SetToken` call already has all the necessary info, no?
So it would just need to additionally save the secretPath in a map on the vault client instance. Or in other words, every time you call `SetToken` for a path, you also have to call `SetDeletePaths` with the same path which results in a rather bad experience.

Kidswiss: I redefined the vault interface and implemented the deletion logic changes we discussed.

pkg/vault/client.go
```diff
@@ -85,16 +119,138 @@ func (b *BankVaultClient) SetToken(secretPath, token string, log logr.Logger) er
...
+				return err
+			}
+		case synv1alpha1.DeletePolicy:
+			b.log.Info("destroying secret", "secret", removeSecret.Path)
+			err := b.destroyToken(path.Join(b.secretEngine, "metadata", removeSecret.Path, secret), versions)
+			if err != nil {
+				return err
+			}
+		default:
+			continue
```
srueg, line 190: Should we return an error here?
Silently ignoring seems dangerous.

go.mod
```diff
@@ -5,14 +5,22 @@ go 1.13
...
+	github.com/docker/go v1.5.1-1 // indirect
 	github.com/go-logr/logr v0.1.0
+	github.com/golangplus/testing v0.0.0-20180327235837-af21d9c3145e
+	github.com/gorilla/mux v1.7.2
 	github.com/hashicorp/vault/api v1.0.4
 	github.com/icza/gox v0.0.0-20200525134802-370c390b446f
 	github.com/imdario/mergo v0.3.9 // indirect
 	github.com/operator-framework/operator-sdk v0.17.1
+	github.com/shortstories/logrusr v0.0.0-20191107062355-5bdb2130acc3
+	github.com/sirupsen/logrus v1.5.0
```
srueg, line 18: Where do these logging dependencies come from?

pkg/vault/client.go
Kidswiss: How should we handle the error? Simply logging it? If we just return it until the reconcile, we'll generate an infinity reconcile loop...

go.mod
Kidswiss: Some testing with other logr interfaces, as I was a bit fed up with json logging for local testing... Forgot to `go mod tidy`

pkg/helpers/crd.go
```diff
@@ -102,3 +123,78 @@ func (s SecretSortList) Less(i, j int) bool {
...
+	if SliceContainsString(instance.GetFinalizers(), finalizerName) && !protected {
+
+		controllerutil.RemoveFinalizer(instance, finalizerName)
+		obj, ok := instance.(runtime.Object)
+		if !ok {
+			return false, fmt.Errorf("can't convert object to runtime.instance")
+		}
+
+		// if this fails the finalizer will remain, effectively not deleting the CR
+		err = client.Update(context.TODO(), obj)
```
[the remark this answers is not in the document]
Kidswiss: I don't see the benefit of that. If the finalizers are removed, the CR is effectively deleted, so the reconcile function should only cleanup and return.

Also if the reconcile function should do that there's code duplication and stuff that might get lost when implementing a new controller.

pkg/controller/gitrepo/gitrepo_reconcile.go
```diff
@@ -46,71 +39,33 @@ func (r *ReconcileGitRepo) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			reqLogger.Info("creating git repo", manager.SecretEndpointName, repo.FullURL())
 			err := repo.Create()
 			if err != nil {
 				return r.handleRepoError(err, instance, repo)
 			}
 			reqLogger.Info("successfully created the repository")
 		}
 
+		deleted, err := helpers.HandleDeletion(instance, finalizerName, r.client)
+		if err != nil || deleted {
```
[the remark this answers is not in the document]
Kidswiss: Right, I'll split that if.

pkg/vault/client.go
srueg: Yes, return it. The exponential backoff will make sure it's retried with less frequency.

pkg/helpers/crd.go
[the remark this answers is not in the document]
srueg: hmm true. The object should be gone if the update was successful.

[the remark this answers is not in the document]
Kidswiss: So now if the finalizer is removed it will return true so the reconcile knows that it should update the object and remove the external resources. Thus deleting it (if all finalizers are gone).

pkg/controller/cluster/cluster_reconcile.go
```diff
@@ -105,6 +139,9 @@ func (r *ReconcileCluster) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 
 	helpers.AddTenantLabel(&instance.ObjectMeta, instance.Spec.TenantRef.Name)
+	helpers.AddDeletionProtection(instance)
```
srueg, line 142: What happens if we did the cleanup and removed the finalizer but there's another finalizer still set?
In that case, our finalizer will be added again here and the cleanup will run again on next reconcile. Maybe the `helpers.HandleDeletion` needs to differentiate between: 1. object not deleted, 2. object deleted and our finalizer is set and 3. object deleted but our finalizer is not set.

Kidswiss: It will now return a struct, which indicates, if the object is deleted and if a finalizer was removed.

So now we can only trigger the actual cleanups once the finalizers are actually removed.

Kidswiss:
Tested it on synfra-dev:

* the operator doesn't have permission to delete/archive the repositories. Gitlab returns a 403 permission denied...
* somehow the creation of the `common.yml` and the cluster yml files broke, they were not created on the dev instance. The operator didn't log any errors or any indication that it tried to create the files.

Kidswiss:
TODO: handle the case where no files are in the repository. This currently just returns.

Kidswiss:
I was able to fix the issue with the empty repository.

Kidswiss:
Fixing this issue and testing against out dev environment shows, that the vault permissions for the operator aren't yet correct.

srueg:
LGTM.

Can you please rebase/merge `master`? Maybe this also fixes the stuck GitHub Actions...

Kidswiss:
Sure I'm squashing the commits anyways before merging.

Issue #83, opened 2020-07-16
Fix issues with gitlab.com


tobru:
Managing git repos on gitlab.com used to work, but now fails with the following error:

```
{"level":"error","ts":1594726227.2132478,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"gitrepo-controller","request":"lieutenant/c-divine-dew-5824","error":"status code was 503","...}
```

As this breaks the [getting started guide](https://syn.tools/syn/getting-started.html) we should investigate and eventually fix the issue.

Kidswiss:
The issue is with the git implementation detection and CloudFlare.

To detect if it's a gitlab instance the operator won't query the API as every API call already requires valid credentials. So it checks if the gitlab login page exists.

That's where CloudFlare's browser detection comes in. It detects that the operator is not a browser and throws the 503.  It can easily be reproduced via `curl`:

```
➜ curl https://gitlab.com/users/sign_in -I
HTTP/2 503
date: Wed, 22 Jul 2020 11:29:48 GMT
content-type: text/html; charset=UTF-8
x-frame-options: SAMEORIGIN
set-cookie: __cfduid=d658dd27d35c485195fb96ca79a9329ae1595417388; expires=Fri, 21-Aug-20 11:29:48 GMT; path=/; domain=.gitlab.com; HttpOnly; SameSite=Lax; Secure
cache-control: private, max-age=0, no-store, no-cache, must-revalidate, post-check=0, pre-check=0
expires: Thu, 01 Jan 1970 00:00:01 GMT
cf-request-id: 0417e27e6c0000cc4ae424b200000001
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 5b6cd3771aeecc4a-ZRH
```

PR #88, opened 2020-07-22
Fix gitlab.com detection due to CloudFlare

Kidswiss:
The check to determine if it's a gitlab instance or not will query the
`/users/sign_in` path. If that path is not found it's not a GitLab
instance. Gitlab.com is using some CloudFlare browser detection on that
path, which works very well and blocks that request from the operator.

Now the operator will determine that it's a GitLab instance if the
endpoint contains `gitlab.com`.

Closes: #83


pkg/git/gitlab/gitlab.go
```diff
@@ -230,23 +230,27 @@ func (g *Gitlab) FullURL() *url.URL {
...
 		Timeout: 10 * time.Second,
 	}
 
-	gitlabURL := URL.Scheme + "://" + URL.Host + "/users/sign_in'"
+	gitlabURL := URL.Scheme + "://" + URL.Host + "/users/sign_in"
 
 	resp, err := httpClient.Get(gitlabURL)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("can't query gitlab login page: %v", err)
```
srueg, line 245:
```suggestion
		return false, fmt.Errorf("can't query gitlab login page: %w", err)
```

pkg/git/gitlab/gitlab.go
```diff
@@ -230,23 +230,27 @@ func (g *Gitlab) FullURL() *url.URL {
 // it will return false.
 func (g *Gitlab) IsType(URL *url.URL) (bool, error) {
 
+	if strings.Contains(URL.Host, "gitlab.com") {
```
srueg, line 233: I really don't like using such "magic strings".
We could use the [/version](https://docs.gitlab.com/ee/api/version.html) endpoint?

Kidswiss: It's not available on gitlab.com:
```
➜ curl https://gitlab.com/version
<html><body>You are being <a href="https://gitlab.com/users/sign_in">redirected</a>.</body></html>%    
```

pkg/git/gitlab/gitlab.go
srueg: What's the reason to stick with `%v` here?

pkg/git/gitlab/gitlab.go
srueg: Another idea would be to just return `true` and think about this problem again once we actually implement other Git providers.

I'm not even sure anymore it makes sense to auto discover the git provider like this. We might as well just include another field in the credentials secret to state which git provider it is. This just seems very brittle and probably causes more errors and confusion than it helps.

What do you think?

Kidswiss: Fine by me.

Until we implement the change correclty I'll do the `return true` thing.

pkg/git/gitlab/gitlab.go
Kidswiss: Habit. Changed it to `%w`.

srueg:
Now the test `TestGitlab_IsType/is_not_Gitlab` fails 🙈

Kidswiss:
I completely removed the `TestGitlab_IsType` now as the function doesn't really do anything anymore.

We should open a new issue to outline how we want to handle the git type definition later on. I don't think that has much prio now as we're only supporting GitLab anyway.

PR #116, opened 2020-10-19
Add configurable revisions for git repositories

corvus-ch:
https://github.com/projectsyn/commodore/issues/198 and https://github.com/projectsyn/commodore/issues/171 asks for a way to store git revisions. After discussing this with @simu  and @srueg, I opted to:

* Add a git URL for the global config repository to the tenant.
* Add revision fields for the global and the tenant config repos to both tenants and clusters.

For easier review, I have split the addition of the new fields and update of the generated resources into two separate commits. 

## Checklist

### DIFF
--- a/pkg/apis/syn/v1alpha1/cluster_types.go
+++ b/pkg/apis/syn/v1alpha1/cluster_types.go
@@ -20,6 +20,10 @@ type ClusterSpec struct {
 	GitRepoTemplate *GitRepoTemplate `json:"gitRepoTemplate,omitempty"`
 	// TenantRef reference to Tenant object the cluster belongs to.
 	TenantRef corev1.LocalObjectReference `json:"tenantRef,omitempty"`
+	// TenantGitRepoRevision allows to configure the revision of the tenant configuration to use. It can be any git tree-ish reference. The revision from the tenant will be inherited if left empty.
+	TenantGitRepoRevision string `json:"tenantGitRepoRevision,omitempty"`
+	// GlobalGitRepoRevision allows to configure the revision of the global configuration to use. It can be any git tree-ish reference. The revision from the tenant will be inherited if left empty.
+	GlobalGitRepoRevision string `json:"globalGitRepoRevision,omitempty"`
 	// TokenLifetime set the token lifetime
 	TokenLifeTime string `json:"tokenLifeTime,omitempty"`
 	// Facts are key/value pairs for statically configured facts
--- a/pkg/apis/syn/v1alpha1/tenant_types.go
+++ b/pkg/apis/syn/v1alpha1/tenant_types.go
@@ -10,6 +10,12 @@ type TenantSpec struct {
 	DisplayName string `json:"displayName,omitempty"`
 	// GitRepoURL git repository storing the tenant configuration. If this is set, no gitRepoTemplate is needed.
 	GitRepoURL string `json:"gitRepoURL,omitempty"`
+	// GitRepoRevision allows to configure the revision of the tenant configuration to use. It can be any git tree-ish reference. Defaults to HEAD if left empty.
+	GitRepoRevision string `json:"gitRepoRevision,omitempty"`
+	// GlobalGitRepoURL git repository storing the global configuration.
+	GlobalGitRepoURL string `json:"globalGitRepoURL,omitempty"`
+	// GlobalGitRepoRevision allows to configure the revision of the global configuration to use. It can be any git tree-ish reference. Defaults to HEAD if left empty.
+	GlobalGitRepoRevision string `json:"globalGitRepoRevision,omitempty"`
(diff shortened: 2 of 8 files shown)


pkg/apis/syn/v1alpha1/tenant_types.go
```diff
@@ -10,6 +10,12 @@ type TenantSpec struct {
 	DisplayName string `json:"displayName,omitempty"`
 	// GitRepoURL git repository storing the tenant configuration. If this is set, no gitRepoTemplate is needed.
 	GitRepoURL string `json:"gitRepoURL,omitempty"`
+	// GitRepoRev allows to configure the revision of the tenant configuration to use. It can be any git tree-ish reference. Defaults to HEAD if left empty.
+	GitRepoRev string `json:"gitRepoRev,omitempty"`
```
simu, line 14: Standardize on `Rev` or `Revision`, but don't use one variant here and the other for the cluster CR.

simu (changes requested):

pkg/apis/syn/v1alpha1/tenant_types.go
corvus-ch: Seems like I got lost somewhere in my decision making. That was certainly not my intention.

srueg: I'd rather go with `GitRepoRevision` since it's a lot clearer IMHO.

srueg:
Seems like another `make generate` is necessary after the changed godoc?

simu (approved):
LGTM overall, although I agree with @srueg that `Revision` is probably clearer for the field names.

srueg:
How do you feel about adding a configurable default (i.e. env var) to set a default for the tenant's `globalGitRepoURL`?
It would be rather cumbersome to define the (currently) same URL on all tenants.

pkg/apis/syn/v1alpha1/tenant_types.go
corvus-ch: I took inspiration from `TenantRef`. But here you go, changed.

pkg/apis/syn/v1alpha1/cluster_types.go
```diff
@@ -20,6 +20,10 @@ type ClusterSpec struct {
 	GitRepoTemplate *GitRepoTemplate `json:"gitRepoTemplate,omitempty"`
 	// TenantRef reference to Tenant object the cluster belongs to.
 	TenantRef corev1.LocalObjectReference `json:"tenantRef,omitempty"`
+	// TenantGitRepoRevision allows to configure the revision of the tenant configuration to use. It can be any git tree-ish reference. The revision from the tenant will be inherited if left empty.
+	TenantGitRepoRevision string `json:"tenantGitRepoRevision,omitempty"`
+	// GlobalRepoRev allows to configure the revision of the global configuration to use. It can be any git tree-ish reference. The revision from the tenant will be inherited if left empty.
```
srueg, line 25:
```suggestion
	// GlobalGitRepoRevision allows to configure the revision of the global configuration to use. It can be any git tree-ish reference. The revision from the tenant will be inherited if left empty.
```

pkg/apis/syn/v1alpha1/tenant_types.go
```diff
@@ -10,6 +10,12 @@ type TenantSpec struct {
 	DisplayName string `json:"displayName,omitempty"`
 	// GitRepoURL git repository storing the tenant configuration. If this is set, no gitRepoTemplate is needed.
 	GitRepoURL string `json:"gitRepoURL,omitempty"`
+	// GitRepoRevision allows to configure the revision of the tenant configuration to use. It can be any git tree-ish reference. Defaults to HEAD if left empty.
+	GitRepoRevision string `json:"gitRepoRevision,omitempty"`
+	// GlobalRepoURL git repository storing the global configuration.
+	GlobalRepoURL string `json:"globalGitRepoURL"`
+	// GlobalRepoRev allows to configure the revision of the global configuration to use. It can be any git tree-ish reference. Defaults to HEAD if left empty.
+	GlobalRepoRev string `json:"globalGitRepoRevision,omitempty"`
```
srueg, line 18:
```suggestion
	GlobalGitRepoRevision string `json:"globalGitRepoRevision,omitempty"`
```

srueg (changes requested):

srueg (approved):

srueg (changes requested):
As discussed:
Let's implement a defaulting for the `GlobalGitRepoURL`.

corvus-ch:
The environment variable `DEFAULT_GLOBAL_GIT_REPO_URL` can now be used to define a default. If the field is empty, the operate will populate it.

srueg (approved):
Please add the new env var to the configuration reference (docs)

srueg (approved):

Issue #80, opened 2020-06-24
Rethink Reconcile Handling


Kidswiss:
At first the reconcile functions were rather simple and the functionality was added directly there.

But we're now at a point where this doesn't scale very well. So I'd like to introduce some more structure to the reconcile functions. We have to to things in certain orders, as they could block or affect other steps during the reconcile. All that logic currently resides inside the respective reconcile loops. This makes it very hard to introduce new functionality that has to be done in certain orders. Also there's a lot of repetition as each reconcile has to do various steps that are common for all the reconcile loops (adding certain labels, or writing back the manipulated CR to the API, etc.).

With some inspiration from https://crossplane.io/docs/master/contributing/services_developer_guide.html I'd like to propose some changes:

We'll define an interface(s) that exposes functions for:
* Fetching the CR
* Checking the CR against all mandatory fields (labels etc.)
* Comparing the state of the external resource (Vault, Git, etc) against the CR definition
* Triggering the state change for the external resource
* Reflect the actual state in the CR
* Write the CR back to the K8s API
* Other things and maybe some future functionality

These functions are roughly in two categories: determine the state and apply the state and may be split from each other. 

Then we'd need some controller, that will go through these functions and determine what actions have to be taken. This sounds like something that could easily be modelled by a FSM (finite state machine). Where it will transition through the various possible states until it reaches the final state where everything is in sync.

This has various benefits:
* The workflow is exactly the same for all our CRs defined by the FSM
* Adding new CRs to the operator is as easy as implementing the interface(s) and feeding to the FSM
* New states can be added easily, golang example for a flexible FSM: https://levelup.gitconnected.com/implement-a-finite-state-machine-in-golang-f0438b6bc0a8

Kidswiss:
After a detailed call with @corvus-ch my idea to use a full blown FSM may be overkill. His suggestions:

We can use a weighted list, where all the various, small steps are added. Each of these steps has a single responsibility, like `addLabel`, `CreateGit`, `RemoveSecrets`, etc.

In this new design, the reconcile function only fetches the necessary information from the cluster to create a state. The list of steps will be filtered according to the state, to get a smaller list of steps that need to be executed. This makes it very easy to extend the reconcile logic, as well as adding the steps at the right point.

Kidswiss:
Interesting article by redhat, on best practices for reconciling: https://www.openshift.com/blog/kubernetes-operators-best-practices

Kidswiss:
[...]

Issue #45, opened 2020-04-22
Set OwnerReference of Clusters to Tenant


srueg:
Set an [ownerReference](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents) on newly created clusters to the respective tenant object.

Just as we do for GitRepo objects.

The result is that cluster objects will then automatically be deleted if the owner (tenant) is deleted. This can be good or bad, maybe an opt-out mechanism (i.e. annotation) needs to be available?

tobru:
What about a finalizer on the Tenant which blocks deletion until all references are gone?

srueg:
That is done automatically with ownerReferences. If you delete an owner with delete propagation "foreground" it adds a finalizer which is only removed once all owner objects are deleted.

PR #120, opened 2020-10-21
Refactor reconciler

Kidswiss:
The reconcile functions are now splitted into smaller logical step functions that can be composed.

These step functions all follow the same interface, so they can be composed at will and accept an interface type for the CRD. There are a few exceptions that will cast the interfaces to concrete objects within the function.

Also each controller now strictly only manipulates its own objects. The git repository objects are still created by other controllers, but changes won't be propagated. Instead the git repositories will now collect the information by themselves.

Relates to: #80

## Checklist


pkg/controller/cluster/cluster_controller.go
```diff
@@ -38,10 +38,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
-
 	return c.Watch(&source.Kind{Type: &synv1alpha1.GitRepo{}}, &handler.EnqueueRequestForOwner{
```
srueg, line 41: Until we implemented #45 we could implement our own `handler.EnqueueRequestsFromMapFunc` to match on the `syn.tools/tenant` label.

This also applies to all the `EnqueueRequestForOwner` types of watches.

Kidswiss: I implemented #45 while debugging the reconcile lag for the tenant objects.

Kidswiss: Reading into the docs of `handler.EnqueueRequestsFromMapFunc` this could help solve the issues with the git repository lag I think.

Kidswiss:
`handler.EnqueueRequestsFromMapFunc` helps quite a lot with the issues I had. Now the gitRepositories can watch tenants and clusters for changes and immediately apply the changed files to the repositories.

Right now I'm adding at least some simple unit tests to all the steps that are now single functions. After that the changes should be okay to review.

Kidswiss:
Finished all the new unittests and fixes for the old ones.

This can now be reviewed. Not sure if I should split this up in some smaller PRs though.

corvus-ch (changes requested):
* Keep an eye on the scope of assertions
* Consider to use loops for the pipelines
* Consider to use maps for test cases

pkg/pipeline/common_steps_test.go
```diff
@@ -0,0 +1,359 @@
...
+	"testing"
+	"time"
+
+	"github.com/projectsyn/lieutenant-operator/pkg/apis"
+	synv1alpha1 "github.com/projectsyn/lieutenant-operator/pkg/apis/syn/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubectl/pkg/scheme"
```
srueg, line 15: I think this should be `"k8s.io/client-go/kubernetes/scheme"`, no?

I really hope we don't have to pull in the `k8s.io/kubectl` dependency.

pkg/pipeline/common_steps.go
```diff
@@ -0,0 +1,166 @@
...
+		err = data.Client.Status().Update(context.TODO(), rtObj)
+	}
+	return ExecutionResult{Abort: true, Err: err}
+}
+
+func wrapError(name string, err error) error {
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("step %s failed: %w", name, err)
```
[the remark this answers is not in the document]
srueg: That's what the `%w` format string actually does: https://golang.org/pkg/fmt/#Errorf

> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand.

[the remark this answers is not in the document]
corvus-ch: That is a nice one. TIL.

pkg/pipeline/common_steps_test.go
Kidswiss: Looks like the vscode autoimporter grabbed some random dependency again.

pkg/pipeline/pipelines.go
```diff
@@ -0,0 +1,73 @@
+package pipeline
+
+// Function defines the general form of a pipeline function.
+type Function func(PipelineObject, *ExecutionContext) ExecutionResult
+
+func ReconcileTenant(obj PipelineObject, data *ExecutionContext) error {
+
+	result := tenantSpecificSteps(obj, data)
```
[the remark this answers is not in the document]
Kidswiss: I like that idea! Good middle ground between using a queue and hand wiring everything like I did here.

pkg/pipeline/cluster_steps_test.go
```diff
@@ -0,0 +1,266 @@
...
+				t.Errorf("createClusterRBAC() = had error: %v", got.Err)
+			}
+		})
+
+		roleBinding := &rbacv1.RoleBinding{}
+		serviceAccount := &corev1.ServiceAccount{}
+
+		namespacedName := types.NamespacedName{Name: tt.args.obj.Name, Namespace: tt.args.obj.Namespace}
+
+		assert.NoError(t, client.Get(context.TODO(), namespacedName, roleBinding))
```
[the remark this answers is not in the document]
Kidswiss: Most of the asserts not within `t.Run()` are probably not intended to be there. Probably a side effect of "see the forest for the trees" I got while creating all those tests.

pkg/pipeline/cluster.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+const (
+	clusterClassContent = `classes:
+- %s.%s
+`
+)
+
+func clusterSpecificSteps(obj PipelineObject, data *ExecutionContext) ExecutionResult {
+	result := createClusterRBAC(obj, data)
```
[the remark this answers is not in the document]
Kidswiss: Obsolete now after I implemented a generic `RunPipeline` func that accepts a map of steps. It's implemented there in that way.

pkg/controller/gitrepo/gitrepo_controller.go
```diff
@@ -33,13 +34,37 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// Watch for changes to primary resource GitRepo
+	// Watch for changes to primary resource Tenant
```
srueg, line 37: The primary resource *is* the GitRepo, no?

pkg/pipeline/common.go
```diff
@@ -0,0 +1,44 @@
...
+
+func common(obj PipelineObject, data *ExecutionContext) ExecutionResult {
+
+	result := handleDeletion(obj.GetObjectMeta(), data)
+	if result.Abort || result.Err != nil {
+		result.Err = wrapError("deletion", result.Err)
+		return result
+	}
+
+	result = addTenantLabel(obj, data)
```
srueg, line 19: Is this really a common step? Currently the tenants themselves don't have a tenant label. It should probably happen together with `setTenantOwner()`?

pkg/pipeline/common_steps.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	err := data.Client.Update(context.TODO(), rtObj)
+	if err != nil {
+		return ExecutionResult{Err: err}
+	}
+
+	// Updating the status if either there were changes or the object is deleted will
+	// lead to some race conditions. By checking first we can avoid them.
+	if resourceVersion == obj.GetObjectMeta().GetResourceVersion() && !data.Deleted {
```
srueg, line 91: Might be more clear:
```suggestion
	if resourceVersion == rtObj.GetObjectMeta().GetResourceVersion() && !data.Deleted {
```
The call to `Update()` will always increment the `ResourceVersion` so the status is never updated?

pkg/pipeline/cluster_steps_test.go
```diff
@@ -0,0 +1,266 @@
...
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+func Test_createClusterRBAC(t *testing.T) {
+	type args struct {
```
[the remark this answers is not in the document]
Kidswiss: I moved them outside the test function and indeed the readability improved quite a bit.

I just moved them. I did not invest time to find cases that can be re-used. I feel like that's more in the scope of its own ticket.

pkg/pipeline/common.go
Kidswiss: Right, I'll move it to gitrepo and cluster specific steps.

pkg/pipeline/common_steps.go
Kidswiss: According to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency

it's only increased when the object is modified.

Kidswiss: > Might be more clear:

The `rtObj runtime.Object` does not expose the ObjectMetada. So it can't be used in there.

pkg/pipeline/cluster_steps_test.go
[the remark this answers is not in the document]
corvus-ch: Mentioning re usability was a general argument. I did not intend to suggest it applies here. So fine by me.

pkg/pipeline/common_steps.go
srueg: Ah, I see. Well then 👌

corvus-ch (changes requested):
All those excessive empty lines makes the code visually appear restless. This is irritating me. I have left some suggestions but then stopped as it started to become excessive.

I suggest to at least drop empty lists after `{` and before `}`.

pkg/pipeline/pipelines.go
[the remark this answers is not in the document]
corvus-ch: To be honest, I had a more elaborate solution in mind. That one probably was overkill. The implementation we have no, is exactly what I was aiming for.

pkg/pipeline/cluster.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func clusterSpecificSteps(obj PipelineObject, data *ExecutionContext) ExecutionResult {
+
+	steps := map[string]func(PipelineObject, *ExecutionContext) ExecutionResult{
+		"create cluster RBAC":    createClusterRBAC,
+		"deletion check":         checkIfDeleted,
+		"set bootstrap token":    setBootstrapToken,
+		"create or update vault": createOrUpdateVault,
+		"delete vault entries":   createOrUpdateVault,
```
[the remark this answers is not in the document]
Kidswiss: Good catch. Fixed

Kidswiss:
I took almost all afternoon to find some very weird and racy issues with the integration tests. After some time I found that our implementation with loops to apply the steps was deeply flawed when using maps. They are not ordered in Go. So the order of the steps was random each time.

I changed it to a slice and struct now everything seems to be working properly again.

srueg (approved):

pkg/pipeline/git_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestCreateOrUpdateGitRepo(t *testing.T) {
+	type args struct {
```
corvus-ch, line 20: There is no need to define a named type when it is only used one. You can define the variable along the data structure.

pkg/pipeline/git_test.go
```diff
@@ -0,0 +1,349 @@
...
+		obj       *synv1alpha1.Cluster
+		template  *synv1alpha1.GitRepoTemplate
+		tenantRef v1.LocalObjectReference
+	}
+	tests := []struct {
```
corvus-ch, line 25: Please move this outside of the test function. And why is this an list and not a dict?

Does also apply to the other test cases within this file.

pkg/pipeline/git_test.go
```diff
@@ -0,0 +1,349 @@
...
+			continue
+		}
+
+		namespacedName := types.NamespacedName{
+			Name:      tt.args.obj.GetName(),
+			Namespace: tt.args.obj.GetNamespace(),
+		}
+
+		checkRepo := &synv1alpha1.GitRepo{}
+		assert.NoError(t, cl.Get(context.TODO(), namespacedName, checkRepo))
```
corvus-ch, line 93: Why are this assertions not within `t.Run`? They will make the `TestCreateOrUpdateGitRepo` fail instead of `TestCreateOrUpdateGitRepo/<name>`.

pkg/pipeline/pipelines.go
```diff
@@ -0,0 +1,53 @@
...
+		{Name: "add tenant label", F: addTenantLabel},
+		{Name: "common", F: common},
+	}
+
+	return RunPipeline(obj, data, steps)
+}
+
+func RunPipeline(obj PipelineObject, data *ExecutionContext, steps []Step) error {
+	for _, step := range steps {
+		if r := step.F(obj, data); resultNotOK(r) {
```
corvus-ch, line 47: Do we still need the `resultNotOK` and `wrapError` abstractions? If I am not mistaken, they are only used here. If this is the case, drop them and do it within this function.

corvus-ch (changes requested):
The code style of one test file completely goes against what was established within the others. Please align that one with the others.

pkg/pipeline/git_test.go
chrisglass: This is a pattern I've often seen with golang tests, I'm not sure I understand your concern. Why should this be a dict?

corvus-ch: Using a list, requires you to have a name property within the struct. Using a map does give you that for free. On top of that, it ensures uniqueness of test names. I also find it easier to visually identify where a test case start and I immediately get an idea about its purpose.

I do not recall the URL but I have read an article that suggest using maps as a best practice.

chrisglass: I see, thanks for the explanation. 

Since it is functionally equivalent, I don't feel strongly about changing the current solution however, personally.

pkg/pipeline/git_test.go
Kidswiss: Looks like I've missed this file when changing the tests.

FWIW: anonymous variables in golang are quite a pain, here's a very nice example: https://stackoverflow.com/a/54126129. That's the reasoning behind this definition.

corvus-ch: I see, but when defining the variable and the data structure within the same statement, this is not an issue. At least it never was for me.

corvus-ch (approved):
LGTM. Thanks for the work.

Disclaimer. I only reviewed coding style and coding structure. I am unable to assert functionality of the code.

Kidswiss:
Thanks @corvus-ch 

From the automated reconcile tests and my manual testing the functionality should be okay.

PR #125, opened 2020-11-03
Unarchive project if it's being recreated

srueg:
## Checklist


Kidswiss:
LGTM

Kidswiss (approved):

pkg/git/gitlab/gitlab.go
```diff
@@ -197,6 +197,12 @@ func (g *Gitlab) getProject() error {
 	if err != nil {
 		return err
 	}
+	if project.Archived {
```
corvus-ch, line 200: This is dangerous. A get operation suggests to be read only. Having the unarchive operation within this function, could result in unexpected behavior.

I would prefer to have this function returning an error and the calling code taking care of calling unarchive if it makes sense.

corvus-ch (changes requested):
This change goes down a dangerous route. Please consider to do this differently.

pkg/git/gitlab/gitlab.go
srueg: You're absolutely right. Thinking about it again, I don't think it's a good idea to un-archive a repo. I guess the current behavior of failing if the repo already exists in an archived state is ok.

PR #126, opened 2020-11-03
Don't set git repo URL for unmanaged repos

srueg:
Otherwise the URL will be overwritten with an empty string.

## Checklist


corvus-ch (changes requested):
Cover this with a test please.

corvus-ch (approved):

PR #130, opened 2020-11-06
Avoid reconcile loop

Kidswiss:
The previous implementation caused the CRs to be updated with each
reconcile. Although no actual changes were applied the ResourceVersion
increased every time. This lead to the status not being set correctly.
And of course excessive reconciling.

By uptdating the status first and catching conflicts, we can avoid that.
This bug was only triggered for newly created GitRepos.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [unrelease]
+## [Unreleased]
 ### Fixed
 - Update loop for newly created GitRepos ([#130])
 
@@ -133,4 +133,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#124]: https://github.com/projectsyn/lieutenant-operator/pull/124
 [#126]: https://github.com/projectsyn/lieutenant-operator/pull/126
 [#127]: https://github.com/projectsyn/lieutenant-operator/pull/127
-[#120]: https://github.com/projectsyn/lieutenant-operator/pull/130
+[#130]: https://github.com/projectsyn/lieutenant-operator/pull/130
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [unrelease]
+### Fixed
+- Update loop for newly created GitRepos ([#130])
+
 ## [v0.4.1] - 2020-11-05
 ### Changed
 - Redesigned reconcile handling ([#120])
@@ -129,3 +133,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#124]: https://github.com/projectsyn/lieutenant-operator/pull/124
 [#126]: https://github.com/projectsyn/lieutenant-operator/pull/126
 [#127]: https://github.com/projectsyn/lieutenant-operator/pull/127
+[#120]: https://github.com/projectsyn/lieutenant-operator/pull/130
--- a/pkg/controller/tenant/tenant_reconcile.go
+++ b/pkg/controller/tenant/tenant_reconcile.go
@@ -32,5 +32,7 @@ func (r *ReconcileTenant) Reconcile(request reconcile.Request) (reconcile.Result
 		FinalizerName: "",
 	}
 
-	return reconcile.Result{}, pipeline.ReconcileTenant(instance, data)
+	res := pipeline.ReconcileTenant(instance, data)
+
+	return reconcile.Result{Requeue: res.Requeue}, res.Err
 }
--- a/pkg/pipeline/common_steps.go
+++ b/pkg/pipeline/common_steps.go
@@ -95,9 +95,6 @@ func updateObject(obj PipelineObject, data *ExecutionContext) ExecutionResult {
 
 		err := data.Client.Status().Patch(context.TODO(), rtObj, client.MergeFrom(origRtObj))
 		if err != nil {
-			if k8serrors.IsConflict(err) {
-				return ExecutionResult{Abort: true}
-			}
 			return ExecutionResult{Err: err}
 		}
 	}
--- a/pkg/pipeline/common_steps.go
+++ b/pkg/pipeline/common_steps.go
@@ -6,10 +6,14 @@ import (
 	"os"
 	"strconv"
 
+	"errors"
+
 	"github.com/projectsyn/lieutenant-operator/pkg/apis"
 	synv1alpha1 "github.com/projectsyn/lieutenant-operator/pkg/apis/syn/v1alpha1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/equality"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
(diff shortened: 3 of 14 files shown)


pkg/pipeline/common_steps.go
```diff
@@ -79,18 +78,25 @@ func updateObject(obj PipelineObject, data *ExecutionContext) ExecutionResult {
 		}
 	}
 
-	updatedObj := rtObj.DeepCopyObject()
-	err := data.Client.Update(context.TODO(), updatedObj)
+	if !data.Deleted {
+		err := data.Client.Status().Update(context.TODO(), rtObj)
```
srueg, line 82: Wont' we have the same problem here, that `rtObj` will be updated with the current state of it in the cluster?

Or does this call only update the `.status` sub-resource?

Kidswiss: This only updates the subresource.

srueg (approved):

Kidswiss:
It will now compare if there are any changes from fetching the object until its updated.

If there are no changes it won't update. This is to avoid unnecessary `resourceVersion` increases.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [unrelease]
```
srueg, line 7:
```suggestion
## [Unreleased]
```

CHANGELOG.md
```diff
@@ -129,3 +133,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#124]: https://github.com/projectsyn/lieutenant-operator/pull/124
 [#126]: https://github.com/projectsyn/lieutenant-operator/pull/126
 [#127]: https://github.com/projectsyn/lieutenant-operator/pull/127
+[#120]: https://github.com/projectsyn/lieutenant-operator/pull/130
```
srueg, line 136:
```suggestion
[#130]: https://github.com/projectsyn/lieutenant-operator/pull/130
```

pkg/pipeline/common_steps.go
```diff
@@ -69,29 +74,49 @@ func addTenantLabel(obj PipelineObject, data *ExecutionContext) ExecutionResult
...
-			Abort: true,
-			Err:   fmt.Errorf("object ist not a valid runtime.object: %v", obj.GetObjectMeta().GetName()),
+		return ExecutionResult{Err: errors.New("object is not a runtime object")}
+	}
+
+	if !specAndMetaEqual(obj, data.originalObject) {
+		err := data.Client.Update(context.TODO(), rtObj)
+		if err != nil {
+			if k8serrors.IsConflict(err) {
+				return ExecutionResult{Abort: true}
```
srueg, line 86: Wouldn't it be better to also return the error here, to make sure it's immediately being re-queued?

pkg/pipeline/common_steps.go
```diff
@@ -69,29 +74,49 @@ func addTenantLabel(obj PipelineObject, data *ExecutionContext) ExecutionResult
...
 	rtObj, ok := obj.(runtime.Object)
 	if !ok {
-		return ExecutionResult{
-			Abort: true,
-			Err:   fmt.Errorf("object ist not a valid runtime.object: %v", obj.GetObjectMeta().GetName()),
+		return ExecutionResult{Err: errors.New("object is not a runtime object")}
+	}
+
+	if !specAndMetaEqual(obj, data.originalObject) {
+		err := data.Client.Update(context.TODO(), rtObj)
```
srueg, line 83: Like this we'd overwrite the `.status` of `rtObj` again, no?

pkg/pipeline/common_steps.go
```diff
@@ -69,29 +74,49 @@ func addTenantLabel(obj PipelineObject, data *ExecutionContext) ExecutionResult
...
 
-	updatedObj := rtObj.DeepCopyObject()
-	err := data.Client.Update(context.TODO(), updatedObj)
-	if err != nil {
-		return ExecutionResult{Err: err}
+	if !equality.Semantic.DeepEqual(data.originalObject.GetStatus(), obj.GetStatus()) {
+
+		origRtObj := data.originalObject.(runtime.Object)
+
+		err := data.Client.Status().Patch(context.TODO(), rtObj, client.MergeFrom(origRtObj))
```
srueg, line 96: What's the reason for using `Patch()` here?

srueg (changes requested):

pkg/pipeline/common_steps.go
Kidswiss: The `.status` of the `rtObj` is being merged back with the `patch()` below.

pkg/pipeline/common_steps.go
Kidswiss: See above explanation about the merging.

srueg: But that's not the status we updated, but the one we deep copied at the beginning of the reconcile?

pkg/pipeline/common_steps.go
```diff
@@ -69,29 +74,49 @@ func addTenantLabel(obj PipelineObject, data *ExecutionContext) ExecutionResult
...
+		if err != nil {
+			if k8serrors.IsConflict(err) {
```
srueg, line 98: Apparently for `Patch` requests, the concurrency mechanism is skipped, so I don't see how we could end up with a conflict error here anyway.

pkg/pipeline/common_steps.go
Kidswiss: You're right, it gets merged from the wrong instance.

Kidswiss: Explains why it needs two reconciles currently to be applied correctly.

pkg/pipeline/common_steps.go
Kidswiss: If that error happens it should already be re-queued due to the pending change? If we return the error it will be logged, which is the intention to avoid with this `if` statemenet.

pkg/pipeline/common_steps.go
Kidswiss: I just found a weird behaviour. If I rewrite the snipped to look like this:

```
deepCopy := rtObj.DeepCopyObject()

if !specAndMetaEqual(obj, data.originalObject) {
		err := data.Client.Update(context.TODO(), rtObj)
		if err != nil {
			if k8serrors.IsConflict(err) {
				return ExecutionResult{Abort: true}
			}
			return ExecutionResult{Err: err}
		}
	}

	if !equality.Semantic.DeepEqual(data.originalObject.GetStatus(), obj.GetStatus()) {

		origRtObj := data.originalObject.(runtime.Object)

		err := data.Client.Status().Patch(context.TODO(), rtObj, client.MergeFrom(origRtObj))
		if err != nil {
			return ExecutionResult{Err: err}
		}
	}
```

The object WON'T contain the status on the k8s API. Even though stepping through with the debugger shows that `rtObj` and `deepCopy` both contain the status before the `Status().Patch()` call.

If using the snipped as it's currently committed it works as expected. There is something weird going on with that copy.

Kidswiss: This only makes sense if `data.Client.Update(context.TODO(), rtObj)` doesn't update the status of the `rtObj` object.

Kidswiss: Just tested it, `data.Client.Update(context.TODO(), rtObj)` will update everything in `rtObj` to the same state as applied to the cluster except the `Status` field.

But it does not explain why `data.Client.Status().Patch(context.TODO(), rtObj, client.MergeFrom(deepCopy))` won't set the actual status though. As I can verify it being called with the correct values.

Kidswiss: Going through the `Patch()` code it looks like they only get apply IF there's a difference between both given objects. This is not the case in my example with deepCopy. So I propose to leave it as it is now with the `origObj`. This way we don't have to merge the fields ourselves and can leverage the `Patch()` method further.

Reference for the called merge code: https://github.com/evanphx/json-patch/blob/v4.5.0/merge.go#L223

pkg/pipeline/common_steps.go
srueg: Not necessarily. How about we check the returned error for `IsConflict` globally and just return `requeue: true` but no error?

srueg (approved):

pkg/pipeline/common_steps.go
Kidswiss: The pipeline execution currently only passes through the error back to the controller. But we can catch the `IsConflict` in the controller itself and set the requeue.

pkg/controller/tenant/tenant_reconcile.go
```diff
@@ -32,5 +32,13 @@ func (r *ReconcileTenant) Reconcile(request reconcile.Request) (reconcile.Result
 		FinalizerName: "",
 	}
 
-	return reconcile.Result{}, pipeline.ReconcileTenant(instance, data)
+	err = pipeline.ReconcileTenant(instance, data)
+	if err != nil {
+		if errors.IsConflict(err) {
```
srueg, line 37: Does this still work, even though we wrap the error in `pipelines.go:56`?
`fmt.Errorf("step %s failed: %w", step.Name, r.Err)`

srueg: It might need a call to `Unwrap`: `apierror.IsConflict(errors.Unwrap(err))`

Kidswiss: Yeah, no... I changed it.

srueg (approved):

Issue #136, opened 2021-01-19
Introduce a template for tenants


corvus-ch:
## Context

In #110 we introduced a template for Clusters. The same is also valuable for Tenants. This would not only address the request https://github.com/projectsyn/lieutenant-api/issues/89 but would give the ability to set all sorts of defaults. 

For the Cluster template, it made sense to embedd it into the Tenant. For the Tenant template there is no such "parent" object to make use of. Instead a new custom resouces (`TenantTemplate`) must be introduced.

```gherkin
Scenario: Template defined
  Given the TenantTemplate
  """
  apiVersion: syn.tools/v1alpha1
  kind: TenantTemplate
  metadata:
    name: default
  spec:
    clusterTemplate:
      gitRepoTemplate:
        apiSecretRef:
          name: "git-credentials"
        repoName: "{{ .Name }}"
  """
  When I create a tenant
  """
  apiVersion: syn.tools/v1alpha1
  kind: Tenant
  metadata:
    name: my-tenant
  spec:
    displayName: "My Tenant"
    clusterTemplate:
      gitRepoTemplate:
        apiSecretRef:
          name: "my-credentials"
  """
  And the lieutenant operator has reconciled that tenant
  Then the tenant has the value
  """
  apiVersion: syn.tools/v1alpha1
  kind: Tenant
  metadata:
    name: my-tenant
    annotations:
      lieutenant.syn.tools/tenant-template: "default"
  spec:
    displayName: "My Tenant"
    clusterTemplate:
      gitRepoTemplate:
        apiSecretRef:
          name: "my-credentials"
        repoName: "{{ .Name }}"
  """

Scenario: No template defined
  Given no TenantTemplate object exists
  When I create a tenant
  """
  apiVersion: syn.tools/v1alpha1
  kind: Tenant
  metadata:
    name: my-tenant
  spec:
    displayName: "My Tenant"
  """
  And the lieutenant operator has reconciled that tenant
  Then the tenant has the value
  """
  apiVersion: syn.tools/v1alpha1
  kind: Tenant
  metadata:
    name: my-tenant
  spec:
    displayName: "My Tenant"
  """
```

### Out of scope

Support for more than one template. This is left for the future should we have the need for it.

## Alternatives

See https://github.com/projectsyn/lieutenant-api/issues/89.

[...]

PR #138, opened 2021-01-26
Implement TenantTemplate

corvus-ch:
This is the implementation for #136.

## Checklist


pkg/pipeline/tenant_steps.go
```diff
@@ -83,3 +84,25 @@ func setGlobalGitRepoURL(obj PipelineObject, data *ExecutionContext) ExecutionRe
...
+
+func applyTemplateFromTenantTemplate(obj PipelineObject, data *ExecutionContext) ExecutionResult {
+	key := types.NamespacedName{Name: "default", Namespace: obj.GetObjectMeta().GetNamespace()}
+	template := &synv1alpha1.TenantTemplate{}
+	if err := data.Client.Get(context.TODO(), key, template); err != nil {
+		// The absence of a template is not an error.
+		// It simply means that there is nothing to do.
+		// TODO Consider to log this case. Where to get a logger from?
+		return ExecutionResult{}
+	}
```
ccremer, line 96: We can at least ignore the IsNotFound error. Other errors we can hardly ignore like temporary, e.g. network-related errors that may in a later reconciliation be resolved again.
We might consider Conditions here to report back errors, also as a means for the lieutenant-api to respect failed reconciliations (e.g. if there's no Ready condition, throw an API error etc)

pkg/pipeline/tenant_steps_test.go
```diff
@@ -171,3 +172,68 @@ func Test_updateTenantGitRepo(t *testing.T) {
...
+			Spec: synv1alpha1.TenantSpec{
+				DisplayName: "My Tenant",
+			},
+		}
+		tenantOut := &synv1alpha1.Tenant{
+			Spec: synv1alpha1.TenantSpec{
+				DisplayName: "My Tenant",
+				DeletionPolicy: synv1alpha1.DeletePolicy,
+			},
+		}
```
ccremer, line 231: I'm missing a case where the annotation `lieutenant.syn.tools/tenant-template: "default"` gets added

docs/modules/ROOT/pages/how-tos/create-tenant.adoc
```diff
@@ -27,6 +27,16 @@ spec:
...
+== Tenant Templating
+
+Values common to all new tenants, can be put into a template.
+When a new tenant gets created, the operator looks for a
+`TenantTemplate` object named `default`.
+A `TenantTemplate` has the same spec data structure as a `Tenant`.
+If present, the templates values from the template will be merged into the tenant.
+The values set on the tenant will take precedence.
+In addition, a reference to the template will be added as an annotation.
+The annotation is named `lieutenant.syn.tools/tenant-template`, and the value will be the name of the tenant.
```
ccremer, line 39:
```suggestion
== Tenant Templating

Values common to all new tenants, can be put into a `TenantTemplate`.
When a new tenant gets created, the operator looks for a `TenantTemplate` object named `default` in the same namespace as `Tenant`.
A `TenantTemplate` has the same spec data structure as a `Tenant`.
If present, the template's values from the template will be merged into the `Tenant` spec.
The values set on the tenant will take precedence.
In addition, a reference to the template will be added as an annotation.
The annotation is named `lieutenant.syn.tools/tenant-template`, and the value will be the name of the `TenantTemplate`.
```

ccremer (approved):

Kidswiss (approved):
LGTM

PR #140, opened 2021-02-01
Per tenant access to tenant and cluster objects

corvus-ch:
Implements #111

## Checklist

- [x] Keep pull requests small so they can be easily reviewed.
- [x] Update the documentation.
- [x] Update the ./CHANGELOG.md.
- [x] Link this PR to related issues.

PR #141, opened 2021-02-04
Refactor some tests

corvus-ch:
While working on #140, I found that some tests are quite verbose and could be cleaned up. To not bloat the other PR, I extracted those changes to this one.

This PR just moves things around but does not functionally change it.

## Checklist


pkg/controller/tenant/tenant_reconcile_test.go
```diff
@@ -26,6 +24,24 @@ func testSetupClient(objs []runtime.Object) (client.Client, *runtime.Scheme) {
 	return fake.NewFakeClientWithScheme(s, objs...), s
 }
 
+func readObject(t *testing.T, c client.Client, ns types.NamespacedName, obj runtime.Object) {
```
ccremer, line 27:
```suggestion
func fetchObject(t *testing.T, c client.Client, ns types.NamespacedName, obj runtime.Object) {
```

pkg/controller/cluster/cluster_reconcile_test.go
```diff
@@ -27,21 +28,38 @@ import (
 )
 
 // testSetupClient returns a client containing all objects in objs
-func testSetupClient(objs []runtime.Object) (client.Client, *runtime.Scheme) {
+func testSetupClient(objs map[schema.GroupVersion][]runtime.Object) (client.Client, *runtime.Scheme) {
```
ccremer, line 31: why this change to a map? It seems to just bloat the usages

We probably could just load our scheme with
```go
s := runtime.NewScheme()
runtime.Must(synv1alpha1.SchemeBuilder.AddToScheme(s))
runtime.Must(appsv1.SchemeBuilder.AddToScheme(s)) // whatever else needed
```
since I consider launching an operator without our schemes loaded like a compilation error: it should just refuse to start up.

pkg/controller/cluster/cluster_reconcile_test.go
```diff
@@ -452,46 +455,42 @@ func TestReconcileCluster_getServiceAccountToken(t *testing.T) {
 }
 
 func TestReconcileCluster_unmanagedGitRepo(t *testing.T) {
```
ccremer, line 457: As an alternative (or in addition?) we could also name this with the given/when/then syntax. It makes it more clear what the test actually does
```suggestion
func TestReconcileCluster_GivenClusterWithUnmanagedGitRepo_WhenReconcileCluster_Then???(t *testing.T) { // what exactly is expected here?
```

pkg/controller/cluster/cluster_reconcile_test.go
corvus-ch: Objects need to be added along with ther scheme. This becomes relevant in #140. In there, I had the situation that Roles were associated both with `syn.tools` and `rbac.authorization.k8s.io`. The client fails then to load the object due to abiguity.

pkg/controller/cluster/cluster_reconcile_test.go
corvus-ch: If we realy want to do BDD, then we would need to switch to another testing framework. Doing this with gotest, feels awekward. Let us discuss this seperateley and leave this for now.

ccremer (approved):

PR #154, opened 2021-02-16
Iterate all pages in a repository file listing

srueg:
See https://docs.gitlab.com/ce/api/README.html#pagination

## Checklist

- [ ] Update the documentation.

### DIFF
--- a/pkg/git/gitlab/gitlab.go
+++ b/pkg/git/gitlab/gitlab.go
@@ -5,11 +5,13 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/projectsyn/lieutenant-operator/pkg/git/helpers"
-	"github.com/projectsyn/lieutenant-operator/pkg/git/manager"
 	"k8s.io/utils/pointer"
 
+	"github.com/projectsyn/lieutenant-operator/pkg/git/helpers"
+	"github.com/projectsyn/lieutenant-operator/pkg/git/manager"
+
 	"github.com/go-logr/logr"
+
 	synv1alpha1 "github.com/projectsyn/lieutenant-operator/pkg/apis/syn/v1alpha1"
 
 	"github.com/xanzy/go-gitlab"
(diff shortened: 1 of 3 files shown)


pkg/git/gitlab/gitlab.go
```diff
@@ -372,28 +378,39 @@ func (g *Gitlab) CommitTemplateFiles() error {
...
-		if strings.Contains(err.Error(), "Tree Not Found") {
-
-			for name, content := range g.ops.TemplateFiles {
-				files = append(files, manager.CommitFile{
-					FileName: name,
-					Content:  content,
-				})
+	// https://docs.gitlab.com/ce/user/gitlab_com/index.html#pagination-response-headers
+	// For performance reasons, if a query returns more than 10,000 records, GitLab.com doesn’t return the TotalPages header.
+	for resp.TotalPages != resp.CurrentPage {
```
ccremer, line 388: I'm not sure what the right condition for the for-loop should be.
with `resp.NextPage > 0` the API docs actually doesn't say what the value is when it reached the last page. Is it 0 again? or equal to the current page number?
With `TotalPages != CurrentPage` this should work better, but not in the case of specifically "gitlab.com" (when there are 10'000 files I guess)

srueg: It's only files in a specific repo.
Also we might get away with only listing the files in the root dir of a repo. Depending on the template file feature, if it even allows creating files in subdirs.

srueg: Ah sorry, I didn't read the comment. You could try it out against our GitLab for example, if `NextPage` is 0 or not.

ccremer: It seems to be empty, which should probably deserialized to 0 in Go. I just tried it out:
```
curl -i --header "PRIVATE-TOKEN: $PRIVATE_TOKEN" "https://git.vshn.net/api/v4/projects/998/repository/tree?per_page=2&page=10"
HTTP/2 200 
...
content-type: application/json
x-next-page: 
x-page: 10
x-per-page: 2
x-prev-page: 9
x-total: 20
x-total-pages: 10
...
```

corvus-ch (approved):

PR #171, opened 2021-05-27
Increase token validity

cimnine:
Closes SYN-354 by increasing the default token validity to `1d`.

Also:
- Remove dead code
- Fix warning recarding capitalization of error message

## Checklist

- [ ] ~~Link this PR to related issues.~~

### DIFF
--- a/pkg/controller/cluster/cluster_steps.go
+++ b/pkg/controller/cluster/cluster_steps.go
@@ -18,12 +18,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-const (
-	clusterClassContent = `classes:
-- %s.%s
-`
-)
-
 func createClusterRBAC(obj pipeline.Object, data *pipeline.Context) pipeline.Result {
 	objMeta := metav1.ObjectMeta{
 		Name:            obj.GetObjectMeta().GetName(),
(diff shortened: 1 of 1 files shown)


pkg/controller/cluster/cluster_steps.go
```diff
@@ -81,9 +75,9 @@ func setBootstrapToken(obj pipeline.Object, data *pipeline.Context) pipeline.Res
 	return pipeline.Result{}
 }
 
-//newClusterStatus will create a default lifetime of 30 minutes if it wasn't set in the object.
+// newClusterStatus will create a default lifetime of 30 minutes if it wasn't set in the object.
```
ccremer, line 78:
```suggestion
// newClusterStatus will create a default lifetime of 24h if it wasn't set in the object.
```

ccremer (approved):

ccremer:
Somehow there's no change from `go mod tidy`?

pkg/controller/cluster/cluster_steps.go
cimnine: 🤦

cimnine:
> Somehow there's no change from `go mod tidy`?

I had to revert them, as `go mod tidy` does not seem to behave nicely with `go generate`ed code.

PR #175, opened 2021-06-24
Update Operator Framework

cimnine:
This PR updates the operator-framework to 1.9, and also several other dependencies.

## Checklist

- [ ] Link this PR to related issues.


Dockerfile
```diff
@@ -1,18 +1,19 @@
...
 
-RUN make test
-RUN make build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
 
 FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY LICENSE .
+COPY --from=builder /workspace/manager .
+USER 65532:65532
```
ccremer, line 17: I generally would prefer to have a very similar setup to K8up, i.e. let Goreleaser build the image

Makefile
```diff
@@ -1,83 +1,245 @@
...
+# BUNDLE_IMG defines the image:tag used for the bundle.
+# You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
+BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
+
+# Image URL to use all building/pushing image targets
+IMG ?= docker.io/projectsyn/lieutenant-operator:latest
+# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
+CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
```
ccremer, line 43: There are a lot of features only relevant for OLM, which we don't need. To ease maintenance, I recommend to get rid of the unneeded boilerplate.
If it's needed in the future again, we can still build in, however it's highly likely that the code will have changed in meantime anyway until then.

config/rbac/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - role_binding.yaml
+  - leader_election_role.yaml
+  - leader_election_role_binding.yaml
+  # Comment the following 4 lines if you want to disable
+  # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
+  # which protects your /metrics endpoint.
+  - auth_proxy_service.yaml
+  - auth_proxy_role.yaml
+  - auth_proxy_role_binding.yaml
+  - auth_proxy_client_clusterrole.yaml
```
ccremer, line 18: Please double check if there aren't any new default features that we don't need.
Our own deployment of Lieutenant is directly using this Kustomize project and it should work the same.

generate.go
```diff
@@ -1,7 +0,0 @@
-// +build generate
-
-package generate
-
-//go:generate go run github.com/operator-framework/operator-sdk/cmd/operator-sdk generate k8s
-//go:generate go run github.com/operator-framework/operator-sdk/cmd/operator-sdk generate crds
-//go:generate ./create-api-docs.sh
```
ccremer, line 7: You probably didn't intend to remove the API documentation generation for Antora, as it's unrelated to the Operator SDK.
However, that API doc generator is buggy. K8up has a different tool:
* https://github.com/vshn/k8up/blob/master/generate.go#L13
* https://github.com/vshn/k8up/tree/master/docs

I propose to re-add resp. replace the API doc-gen in a separate PR.

ccremer (changes requested):
Thank you managing to do this monster upgrade!

I think we should slim down the SDK-specific boilerplate and align the repository with K8up.
Specifically this affects:
* Makefile -> Remove most of the SDK magic and try to use as much from K8up as possible
* Goreleaser -> Missing completely. Also aligns with K8up. Could also be done in a separate PR, but should be part of the upgrade IMO.
* Kustomize -> It looks like the setup has changed completely over v0.5.x. We officially deploy using Kustomize to our production environment and thus it should work transparently. I noticed there is much that comes with the upgrade. It should be largely compatible with https://github.com/projectsyn/lieutenant-operator/tree/v0.5.3/deploy

hack/boilerplate.go.txt
```diff
@@ -1,15 +0,0 @@
-/*
```
ccremer: I'm not quite sure if you can delete the whole file. I seems to recall a problem when generating the CRDs... maybe that changed though in recent versions of controller-tools or controller-runtime.

In K8up we had to define it here: https://github.com/vshn/k8up/blob/master/generate.go#L5

cimnine: The file is still there, albeit empty.

ccremer:
I have fixed some issues.
When I `k apply -f config/samples/syn_v1alpha1_cluster.yaml` then I observe panics and crashloops. Not sure if that's related to the upgrade or because there's no Vault in the e2e setup.

The integration tests still fail though with Ginkgo

generate.go
```diff
@@ -6,10 +6,4 @@ package main
 //go:generate go run sigs.k8s.io/controller-tools/cmd/controller-gen crd:trivialVersions=true rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=${CRD_ROOT_DIR}/bases crd:crdVersions=v1
 
 // Generate API reference documentation
-////go:generate go run github.com/elastic/crd-ref-docs --source-path=api/v1alpha1 --config=docs/api-gen-config.yaml --renderer=asciidoctor --templates-dir=docs/api-templates --output-path=${CRD_DOCS_REF_PATH}
-
-import (
-	_ "github.com/elastic/crd-ref-docs"
-	_ "sigs.k8s.io/controller-tools/cmd/controller-gen"
-	_ "sigs.k8s.io/kustomize/kustomize/v3"
-)
```
ccremer: Are you really sure this isn't needed when running `make generate`?
(note it might not output anything if the generation failed)

cimnine: yes. they are already covered in `tools.go`.

cimnine: besides, they're not part of the same file in k8up

ccremer (approved):
In the documentation preview, there are a few broken links.
Once those are fixed, it's LGTM

cimnine:
> In the documentation preview, there are a few broken links.

Nice catch! I fixed these now.

PR #181, opened 2021-07-16
Add details about where errors occurred

mhutter:
During troubleshooting issues with Lieutenant, we noticed that some
error messages did not contain enough information to find out which step
of a pipeline failed.

This commit tries to improve the errors reported back to the user by
making sure all errors returned in a `Result` contain some information
about the precise step that triggered it.

## Checklist

- [x] ~Update the documentation.~
- [x] ~Link this PR to related issues.~

### DIFF
--- a/pipeline/steps.go
+++ b/pipeline/steps.go
@@ -89,7 +89,7 @@ func updateObject(obj Object, data *Context) Result {
 				data.Log.V(1).Error(err, "conflict while updating object; requeueing")
 				return Result{Requeue: true}
 			}
-			return Result{Err: err}
+			return Result{Err: fmt.Errorf("update %s/%s: %w", obj.GroupVersionKind().String(), obj.GetName(), err)}
 		}
 	}
 
@@ -105,7 +105,7 @@ func updateObject(obj Object, data *Context) Result {
 				data.Log.V(1).Error(err, "conflict while updating object; requeueing")
 				return Result{Requeue: true}
 			}
-			return Result{Err: err}
+			return Result{Err: fmt.Errorf("update %s/%s status: %w", obj.GroupVersionKind().String(), obj.GetName(), err)}
 		}
 	}
 
--- a/controllers/cluster/bootstrap_token.go
+++ b/controllers/cluster/bootstrap_token.go
@@ -21,7 +21,7 @@ func setBootstrapToken(obj pipeline.Object, data *pipeline.Context) pipeline.Res
 		data.Log.Info("Adding status to Cluster object")
 		err := newClusterStatus(instance)
 		if err != nil {
-			return pipeline.Result{Err: err}
+			return pipeline.Result{Err: fmt.Errorf("setting initial status on cluster: %w", err)}
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


controllers/cluster/rbac.go
```diff
@@ -42,7 +43,7 @@ func createClusterRBAC(obj pipeline.Object, data *pipeline.Context) pipeline.Res
 	}
 	for _, item := range []client.Object{serviceAccount, role, roleBinding} {
 		if err := data.Client.Create(context.TODO(), item); err != nil && !errors.IsAlreadyExists(err) {
-			return pipeline.Result{Err: err}
+			return pipeline.Result{Err: fmt.Errorf("creating %s/%s: %w", item.GetObjectKind().GroupVersionKind().Kind, item.GetName(), err)}
```
ccremer, line 46:
```suggestion
			return pipeline.Result{Err: fmt.Errorf("creating %s '%s': %w", item.GetObjectKind().GroupVersionKind().Kind, item.GetName(), err)}
```

ccremer (approved):

controllers/cluster/rbac.go
ccremer: Actually, the GVK on newly created structs might empty if not explicitly set, the Client figures this out almost via reflection. So the `item.GetObjectKind().GroupVersionKind().Kind` might be empty at this time.
I'm not quite sure though. Did you try it?

pipeline/steps.go
```diff
@@ -95,7 +95,7 @@ func updateObject(obj Object, data *Context) Result {
 			if k8serrors.IsConflict(err) {
 				return Result{Requeue: true}
 			}
-			return Result{Err: err}
+			return Result{Err: fmt.Errorf("update %s/%s status: %w", obj.GroupVersionKind().String(), obj.GetName(), err)}
```
ccremer, line 98:
```suggestion
			return Result{Err: fmt.Errorf("update %s '%s' status: %w", obj.GroupVersionKind().String(), obj.GetName(), err)}
```

pipeline/steps.go
```diff
@@ -84,7 +84,7 @@ func updateObject(obj Object, data *Context) Result {
 			if k8serrors.IsConflict(err) {
 				return Result{Requeue: true}
 			}
-			return Result{Err: err}
+			return Result{Err: fmt.Errorf("update %s/%s: %w", obj.GroupVersionKind().String(), obj.GetName(), err)}
```
ccremer, line 87:
```suggestion
			return Result{Err: fmt.Errorf("update %s '%s': %w", obj.GroupVersionKind().String(), obj.GetName(), err)}
```

corvus-ch (commented):
Thanks for the contribution. I do like the usage of gift wrapping errors and give them a bit more context. However, I dislike the chosen wording style.

Could you rewrite them to use a formal correct sentence? At leas something that contains a verb.

Suggested pattern: `failed to do x: %w`.

mhutter:
> Could you rewrite them to use a formal correct sentence? At leas something that contains a verb.

This would be up to the "user interface". Since the "interface" here is a JSON-formatted log message, and the error messages are only one field of said message, I don't think it's appropriate here.

corvus-ch (approved):

PR #188, opened 2021-08-06
Register Gitlab implemenation

glrf:
During refactor we seem to have lost the import that registered the Gitlab implementation.  This lead to all gitrepo reconciles to fail. This PR will add back this magic import.

I personally really don't like the magic `init` approach, as we have seen these imports can easily be missed when refactoring.
But I would say changing this is out of scope for this bugfix.

## Checklist

### DIFF
--- a/controllers/gitrepo/steps.go
+++ b/controllers/gitrepo/steps.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 
 	synv1alpha1 "github.com/projectsyn/lieutenant-operator/api/v1alpha1"
+
+	// Register Gitrepo implementation - DONOT REMOVE
+	_ "github.com/projectsyn/lieutenant-operator/git"
 	"github.com/projectsyn/lieutenant-operator/git/manager"
 	"github.com/projectsyn/lieutenant-operator/pipeline"
 	"sigs.k8s.io/controller-runtime/pkg/client"


ccremer:
This "magic" is nothing else than [Dependency Inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle) combined with the [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern). The strategy pattern allows different low-level implementations for the same behavior (e.g. GitLab, GitHub, Gitea etc.) while the dependency inversion allows that the high-level code doesn't need to know which concrete implementation it's interacting with, making this pattern a natural choice.

The reason why the register failed is indeed a design oversight: Since the dependency inversion is applied correctly, there's no direct reference to the "../gitlab" package. While the `init()` is very much Go idiomatic, the Go compiler thinks the package is unused and doesn't call the `init()` function at runtime.

You could make this explicit by exporting `init()` and simply call `gitlab.Init()` from `main.go`, this gets rid of the "magic import" 🤷

glrf:
Yes. I do understand how this works, and I get the attempt at decoupling. In my opinion this is a case of YAGNI. I've honestly never seen a good implementation using `Init`. It just hides setup.

ccremer:
> It just hides setup.

Well, you could just name it `Setup()` if that helps?
What would your concrete proposal be?

glrf:
It generally just feels wrong..

I think my main issue actually isn't the init function in of itself, it just showed us that there is an issue IMO. Mainly I don't like the global `implementation` variable. I generally don't expect packages to hold state, it makes tests harder and as soon as we want to do any kind of concurrency we are in for a world of hurt.

 I would probably switch to a `manager` struct. Then I would simplify the package structure. Move the manager struct to the `git` package. Maybe even move the implementation itself there.

Then use dependency injection. Setup the manager struct once, probably in main or similar and inject the manager to where it is needed.

ccremer (approved):

PR #192, opened 2021-08-10
Fix typed nil assignment

glrf:
The newBankVaultClient function returns a typed nil on error. This will cause subsequent calls to panic if we assign it directly to the global variable.

## Checklist


bastjan (approved):
thanks for this one 🚀

PR #194, opened 2021-08-10
Fix vault client logging

bastjan:
Log vault client logs instead of vaults http logs. Note the differing logging interfaces, lovely.

## Checklist


ccremer (approved):
I think it wouldn't hurt to create a small unit test for the flatten func though

bastjan:
> I think it wouldn't hurt to create a small unit test for the flatten func though

You're completely right. I just added one. 😊

PR #196, opened 2021-08-11
Fix error when recreating git repository

glrf:
We fix two bugs in this PR

* When creating (or recreating) a repository we will create a file containing `{delete}` for all deleted class files.
* When adding more than one file we try to create the last file multiple times because of a pointer error

## Checklist


bastjan (commented):
How much work would it be to test the recreation / multiple files creation? 😅

ccremer (approved):

glrf:
All these tests are horribly hacky, but we can catch the original bugs now..

bastjan (approved):
LGTM

PR #197, opened 2021-08-12
Steps and status update debug logs

bastjan:
## Checklist

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -57,8 +57,13 @@ type Step struct {
 }
 
 func RunPipeline(obj Object, data *Context, steps []Step) Result {
-	for _, step := range steps {
-		if r := step.F(obj, data); r.Abort || r.Err != nil {
+	l := data.Log.V(2).WithName("RunPipeline")
+	l.Info("running steps", "steps", stepNames(steps))
+
+	for i, step := range steps {
+		r := step.F(obj, data)
+		l.Info("ran step", "step", step.Name, "result", r, "step_index", i)
+		if r.Abort || r.Err != nil {
 			if r.Err == nil {
 				return Result{Requeue: r.Requeue}
 			}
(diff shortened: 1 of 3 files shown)


pipeline/pipeline.go
```diff
@@ -57,7 +57,12 @@ type Step struct {
...
 func RunPipeline(obj Object, data *Context, steps []Step) Result {
-	for _, step := range steps {
+	l := data.Log.V(2).WithName("RunPipeline")
+	l.Info("running steps", "steps", stepNames(steps))
+
+	for i, step := range steps {
+		r := step.F(obj, data)
+		l.Info("ran step", "step", step.Name, "result", r, "step_index", i)
 		if r := step.F(obj, data); r.Abort || r.Err != nil {
 			if r.Err == nil {
```
glrf, line 67: I think we are now running the step twice?

pipeline/steps_test.go
```diff
@@ -369,3 +372,13 @@ func Test_updateObjectStatus(t *testing.T) {
 	assert.Equal(t, "token-1234", updatedCluster.Status.BootstrapToken.Token)
 	assert.NotEqual(t, updatedCluster.ResourceVersion, cluster.GetResourceVersion())
 }
+
+func addLogger(c *Context) *Context {
+	l, err := zap.NewDevelopment()
+	if err != nil {
+		panic(fmt.Errorf("unexpected error while creating zap logger: %w", err))
```
glrf, line 379: Not sure if I like the panic here. But I guess when logging to stdout this should never happen. So I guess its fine

glrf (changes requested):

pipeline/pipeline.go
bastjan: thanks for catching this one 🤦‍♀️  looks like coffee levels are not high enough today

pipeline/steps_test.go
bastjan: Would need some major test refactoring otherwisem, since contexts are created outside the test functions.

pipeline/pipeline.go
glrf: That's what code reviews are for. I would have hopped that tests would have caught that..

glrf (approved):
LGTM

PR #199, opened 2021-08-12
Fix issue that blocked status updates

glrf:
Update on the main resource does not update the status, but it does fetch the current version of that status from k8s. We currently use the orginal object in this update call, which has its status overwritten.  We need to copy the resource when updating to make sure that we can update both the resource and  its status.

## Checklist

### DIFF
--- a/pipeline/steps.go
+++ b/pipeline/steps.go
@@ -79,9 +79,12 @@ func updateObject(obj Object, data *Context) Result {
 	}
 
 	if !specAndMetaEqual(obj, data.originalObject) {
-		data.Log.V(2).Info("updating object")
-
-		err := data.Client.Update(context.TODO(), rtObj)
+		data.Log.V(1).Info("updating object")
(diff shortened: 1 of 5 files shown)


pipeline/steps.go
```diff
@@ -109,7 +115,7 @@ func updateObject(obj Object, data *Context) Result {
 
 func specAndMetaEqual(a, b Object) bool {
 
-	if !equality.Semantic.DeepEqual(a, b) {
+	if !equality.Semantic.DeepEqual(a.GetMeta(), b.GetMeta()) {
 		return false
 	}
```
ccremer, line 120: The func is named `specAndMetaEqual` but now it only seems to compare the metadata? Is this correct?
Maybe worth testing? (if not already done)

glrf: No the spec is compared below.

ccremer: argh ofc :see_no_evil: :tomato:

ccremer (approved):

PR #201, opened 2021-08-18
Add dynamic fact field to CRD

glrf:
## Checklist

Relates to https://github.com/projectsyn/lieutenant-api/pull/157 and https://github.com/projectsyn/steward/pull/85


controllers/cluster/rbac.go
```diff
@@ -41,7 +41,16 @@ func createClusterRBAC(obj pipeline.Object, data *pipeline.Context) pipeline.Res
...
-		if err := data.Client.Create(context.TODO(), item); err != nil && !errors.IsAlreadyExists(err) {
+		found := item.DeepCopyObject().(client.Object)
+		err := data.Client.Get(context.TODO(), client.ObjectKeyFromObject(item), found)
+		if errors.IsNotFound(err) {
+			if err := data.Client.Create(context.TODO(), item); err != nil && !errors.IsAlreadyExists(err) {
+				return pipeline.Result{Err: err}
+			}
+		} else if err != nil {
+			return pipeline.Result{Err: err}
+		}
```
cimnine, line 52: I suggest to first check if err is not nil, an only then run the errors.IsNotFound check.

controllers/cluster/rbac.go
```diff
@@ -41,7 +41,16 @@ func createClusterRBAC(obj pipeline.Object, data *pipeline.Context) pipeline.Res
...
+		if err := data.Client.Update(context.TODO(), item); err != nil {
```
cimnine, line 53: Maybe you could check if you can get a proper context from the calling function. Otherwise I suggest to deliberately use context.Background() instead of .TODO().

cimnine (approved):
Just some minor would-be-nice's

controllers/cluster/rbac.go
glrf: The `context.TODO()` issues is sadly present all throughout the codebase. We currently simply do not pass an actual context. I might invest an hour or so to fix this.

However I disagree with just switching to `context.Background()`. This is pretty much the one use case for `context.TODO()`

> TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

simu (approved):

PR #202, opened 2021-08-19
Remove all `context.TODO()`

glrf:
## Checklist

Merge after https://github.com/projectsyn/lieutenant-operator/pull/201

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -1,6 +1,7 @@
 package pipeline
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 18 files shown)


cimnine (approved):
LGTM, but I have requested a 2nd opinion from @ccremer

glrf:
> LGTM, but I have requested a 2nd opinion from @ccremer

Chris is on holiday. @corvus-ch could you take a look?

pipeline/pipeline.go
```diff
@@ -42,6 +44,13 @@ type Context struct {
 	Reconciler     reconcile.Reconciler
 }
 
+func (c Context) Context() context.Context {
+	if c.Ctx != nil {
+		return context.Background()
```
bastjan, line 49: I'd prefer to remove this bandaid. Contexts should always be explicitly defined in my mind and not as a side effect.

I guess it is only used in tests?

glrf: Good point. I was unsure whether I should add this. It's just with the way we use the `pipeline.Context` it seems very easy to forget setting the actual context. But I guess we have the same problem anyway if we forget any of the other fields..

bastjan (approved):
LGTM

corvus-ch (approved):

corvus-ch:
Look out for #204. There I have seen some `context.TODO()`.

PR #205, opened 2021-08-20
Fix recreated service account token bug

glrf:
Last PR introduced a bug that removed the service account secret reference on reconcile. This caused kubernetes to recreate the SA token.

## Checklist


bastjan (approved):
LGTM

cimnine (approved):

PR #240, opened 2022-08-03
Remove unnecessary `url.PathEscape` for `GetNamespace`

simu:
The new version of the GitLab library URL-escapes the provided repo path in `GetNamespace`, so we must not escape it ourselves anymore, as we otherwise get a double-URL-escaped value for the request, which then fails.

Also add a test case for creating a project in a subgroup.

Follow-up fix for #217 

## Checklist

### DIFF
--- a/git/gitlab/gitlab.go
+++ b/git/gitlab/gitlab.go
@@ -253,7 +253,7 @@ func (g *Gitlab) New(options manager.RepoOptions) (manager.Repo, error) {
 
 func (g *Gitlab) getNamespaceID() (*int, error) {
 
-	fetchedNamespace, _, err := g.client.Namespaces.GetNamespace(url.PathEscape(g.ops.Path))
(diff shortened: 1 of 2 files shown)


glrf (commented):
I think this looks good to me.

I'm a little confused by the test case. Shouldn't this code call `/api/v4/namespaces/path%2Fto%2Fgroup2`? I guess there is some magic somewhere I'm missing or am I misunderstanding the API?

simu:
> I'm a little confused by the test case. Shouldn't this code call `/api/v4/namespaces/path%2Fto%2Fgroup2`? I guess there is some magic somewhere I'm missing or am I misunderstanding the API?

I'm confused as well, but with `/api/v4/namespaces/path%2Fto%2Fgroup2` the test panics because there's no handler for the requested path. I've not been able to figure out yet where the unescape happens.

glrf:
Aah in Go's `url` package:

```

glrf (approved):

glrf (approved):

PR #272, opened 2024-06-03
Simplify tenant role management; allow tenant SA to `patch` cluster status

bastjan:
Allows commodore CI to set compilation metadata.

Strongly simplifies tenant role management. The PR assumes the created roles, which always had full `controller` owner references set, are fully controller owned. Changes to those roles get overridden by the controller on the next reconcile.


controllers/cluster/steps.go
```diff
@@ -14,7 +14,6 @@ func SpecificSteps(obj pipeline.Object, data *pipeline.Context) pipeline.Result
 		{Name: "delete vault entries", F: vault.HandleVaultDeletion},
 		{Name: "set tenant owner", F: setTenantOwner},
 		{Name: "apply cluster template from tenant", F: applyClusterTemplateFromTenant},
-		{Name: "update Role", F: clusterUpdateRole},
```
bastjan, line 17: This did not update the cluster role. It did update the tenant role. 🙃

controllers/tenant/role.go
```diff
@@ -3,73 +3,59 @@ package tenant
...
+			{
+				APIGroups:     []string{synv1alpha1.GroupVersion.Group},
+				Verbs:         []string{"get"},
+				Resources:     []string{"tenants", "clusters"},
+				ResourceNames: rns,
+			}, {
+				APIGroups:     []string{synv1alpha1.GroupVersion.Group},
+				Verbs:         []string{"get", "update", "patch"},
+				Resources:     []string{"clusters/status"},
+				ResourceNames: rns,
```
simu, line 50: Out of curiosity: do I read this right that this allows `get/update/patch` on `clusters/status` for resource name `t-tenant-1234` in addition to the tenant's clusters?

simu (approved):
LGTM.

controllers/tenant/role.go
bastjan: Right we can do this nicer now 😅

bastjan: fixed in https://github.com/projectsyn/lieutenant-operator/pull/272/commits/1dba9bf8a82b43dad04f37354b3780c8c9ebcf74.

PR #276, opened 2024-06-04
Collect Cluster compilation metadata metrics according to SDD 0031

bastjan:
https://syn.tools/syn/SDDs/0031-component-version-tracking.html#_operator

## Checklist

### DIFF
--- a/metrics/compile_meta_collector.go
+++ b/metrics/compile_meta_collector.go
@@ -1,0 +1,175 @@
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	synv1alpha1 "github.com/projectsyn/lieutenant-operator/api/v1alpha1"
+)
+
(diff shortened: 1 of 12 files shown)


metrics/compile_meta_collector.go
```diff
@@ -0,0 +1,175 @@
...
+		ch <- prometheus.NewInvalidMetric(metaLastCompileDesc, err)
+		ch <- prometheus.NewInvalidMetric(newMetaCommodoreBuildInfoDesc(), err)
+		ch <- prometheus.NewInvalidMetric(metaGlobalDesc, err)
+		ch <- prometheus.NewInvalidMetric(metaTenantDesc, err)
+		ch <- prometheus.NewInvalidMetric(metaPackageDesc, err)
+		ch <- prometheus.NewInvalidMetric(metaInstanceDesc, err)
+	}
+
+	for _, cl := range cls.Items {
+		ch <- prometheus.MustNewConstMetric(
```
simu, line 97: How tricky would it be to only generate the last compile, build info, global and tenant repo metrics when the cluster actually has compile meta information?

simu (approved):

metrics/compile_meta_collector.go
bastjan: Not that hard.

I thought about that but decided that it feels the most natural to me to leave them empty for a query to find clusters without metadata yet. It should not be that many metrics overall with all the component instance and package metrics.

PR #285, opened 2024-07-08
Add CI variable management to GitRepo

bastjan:
This PR:
- Adds a CRD field `accessToken`. If the field is set, an access token will be requested from Gitlab and stored in this field.
- Adds a field `ciVariables` which allow managing (Gitlab) CI variables from secrets.
- Fixed Tenants not reconciled when updating TenantTemplate.

## Checklist


simu (approved):
Implementation LGTM :rocket: . I didn't review the tests in detail.

I was wondering whether we should add info logs for missing optional secrets, but decided that that's probably not necessary.

bastjan:
> I was wondering whether we should add info logs for missing optional secrets, but decided that that's probably not necessary.

Good idea. Added a log. Helps when debugging.

PR #293, opened 2024-07-24
Disentangle random environment variables

HappyTetrahedron:
## Checklist


HappyTetrahedron:
Maybe it'd be better to get rid of `envflags` and rewrite the component to use args instead. Thoughts?

bastjan (approved):

bastjan:
> to use args instead

What kinda args?

I'd say fine as is atm. Using args might be something for lieutenant 2.0.

PR #302, opened 2024-07-31
Sort cluster names in permissions to remove unnecessary updated

bastjan:
This is now an issue with the compile pipelines which changes clusters more often.

## Checklist


HappyTetrahedron (approved):
LGTM. Had to do something similar with the cluster list in the tenant status as well

PR #317, opened 2025-08-13
Prevent the operator from removing the git repo url on transient Gitlab errors

bastjan:
`repoExists()` returned false for every error, not only `ErrNotFound`. If `repo.Create` then also returned an error - which is highly likely because the repo already exists or the transient error still persists - the repo url would be deleted.

## Checklist


HappyTetrahedron (approved):
I approve! I think I ran into this at some point a while ago.

PR #330, opened 2026-02-03
Add support for custom GitLab SSH endpoint

rxbn:
This pull request add support for a custom gitlab ssh endpoint url. I have already verified the changes in our environment.

## Checklist

### DIFF
--- a/git/gitlab/gitlab.go
+++ b/git/gitlab/gitlab.go
@@ -26,9 +26,7 @@ func init() {
 	manager.Register(&Gitlab{})
 }
 
-var (
-	ListItemsPerPage = 100
-)
+var ListItemsPerPage = 100
 
 // Gitlab holds the necessary information to communincate with a Gitlab server.
 // Each Gitlab instance will handle exactly one project.
@@ -43,7 +41,6 @@ type Gitlab struct {
 
 // Create will create a new Gitlab project
 func (g *Gitlab) Create() error {
-
 	nsID, err := g.getNamespaceID()
 	if err != nil {
 		return err
@@ -136,7 +133,6 @@ func (g *Gitlab) updateDisplayName() (bool, error) {
 
 func (g *Gitlab) removeDeployKeys(deleteKeys map[string]synv1alpha1.DeployKey) error {
 	existingKeys, _, err := g.client.DeployKeys.ListProjectDeployKeys(g.project.ID, &gitlab.ListProjectDeployKeysOptions{})
-
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


git/gitlab/gitlab.go
```diff
@@ -227,11 +223,21 @@ func (g *Gitlab) Connect() error {
...
 // FullURL returns the complete url of this git repository
 func (g *Gitlab) FullURL() *url.URL {
-
 	sshURL := *g.ops.URL
 
 	sshURL.Scheme = "ssh"
 	sshURL.User = url.User("git")
+	if g.ops.SSHHost != "" {
+		sshURL.Host = g.ops.SSHHost
+		// If the original URL had no host (for example "git.example.com/foo/bar"),
```
simu: Shouldn't this be "if the original URL had no scheme? Otherwise the example string doesn't really make sense.

```suggestion
		// If the original URL had no scheme (for example "git.example.com/foo/bar"),
```

simu (approved):
Hi @rxbn, thanks for the PR!

The change LGTM, but I'd like to have a second review from @bastjan.

git/gitlab/gitlab.go
rxbn: Good catch, thanks :)

git/manager/manager.go
```diff
@@ -239,5 +249,31 @@ func GetGitClient(ctx context.Context, instance *synv1alpha1.GitRepoTemplate, na
 	err = repo.Connect()
 
 	return repo, hostKeysString, err
+}
+
+func parseSSHEndpoint(raw string) (string, error) {
```
bastjan, line 254: I'd love a quick unit test (possibly table driven) of this private function as there is quite some custom parsing logic there.

bastjan (commented):
LGTM overall, a test for the custom parsing would not hurt.

git/manager/manager.go
rxbn: Thanks @bastjan! I have added a unit test.

git/manager/manager_test.go
```diff
@@ -86,3 +86,50 @@ func TestNewRepo(t *testing.T) {
...
+		},
+		{
+			name: "ssh url with scheme",
+			raw:  "ssh://git@example.com:2222/group/project.git",
+			want: "example.com:2222",
+		},
+		{
+			name: "host without scheme",
+			raw:  "git.example.com/foo/bar",
+			want: "git.example.com/foo/bar",
```
bastjan: Would it make sense for this URL to behave the same way as the one with `ssh://` - stripping the path?

rxbn: You are right, its a host override not a base path. I have patched the func and the test case accordingly. Thanks for the catch 👍🏼

bastjan: Thanks for the fix 😊 LGTM!

PR #340, opened 2026-05-07
Introduce new metric `syn_cluster_dynamic_facts`

simu:
This metric enables us to easily consume Lieutenant cluster dynamic facts through Prometheus queries (e.g. for central dashboards).

We currently don't filter out any dynamic facts, and accept that the metric may have sizeable label values.

Generally speaking, Prometheus should be able to handle bigger label values gracefully, but if the larger label values (currently primarily `status.facts.kubernetesVersion`) turn out to be an issue, we can always drop them with a scrape relabel config (in the `ServiceMonitor` for each Lieutenant instance).

## Checklist

- [ ] Update the documentation.

### DIFF
--- a/metrics/cluster_info_collector.go
+++ b/metrics/cluster_info_collector.go
@@ -31,11 +31,28 @@ var clusterFactsDesc = prometheus.NewDesc(
 	nil,
 )
 
-// commodore build info has dynamic labels
+var clusterDynFactsDesc = prometheus.NewDesc(
+	"syn_lieutenant_cluster_dynamic_facts",
+	"Lieutenant cluster dynamic facts.",
+	[]string{"cluster", "tenant", "display_name"},
+	nil,
+)
+
+// cluster facts has dynamic labels
 func newClusterFactsDesc(lbls ...string) *prometheus.Desc {
 	return prometheus.NewDesc(
 		"syn_lieutenant_cluster_facts",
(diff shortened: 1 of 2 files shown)


metrics/cluster_info_collector.go
```diff
@@ -112,6 +133,39 @@ func clusterFacts(cl synv1alpha1.Cluster, ch chan<- prometheus.Metric) error {
 	return nil
 }
 
+// clusterDynFacts collects the dynamic facts of a cluster and sends them as Prometheus metrics.
+// The keys of the facts are normalized to be valid Prometheus labels.
+// If the first character of a key is an underscore or an invalid character it is replaced with "fact_".
+// If a key is empty it is replaced with "_empty".
+// If a key is in the protected list after normalizing it is prefixed with "orig_".
+// If a key is a duplicate after normalizing it is suffixed with "_<n>" where n is the number of duplicates.
+func clusterDynFacts(cl synv1alpha1.Cluster, ch chan<- prometheus.Metric) error {
```
simu: Not sure if it would make more sense to have a single `clusterFacts()` function which takes a `func (lbls... string) *prometheus.Desc` and either `cl.Spec.Facts` or `cl.Status.Facts` instead of essentially having the same function twice.

bastjan: Makes sense to dry this up a bit I think.

bastjan (approved):

